PR #7, opened 2026-02-12
Fix `KeyMaterial::concatenate` logic and improve `Condition<T>` correctness for `i64`/`u64`

Quant-TheodoreFelix:
## Title

Fix `KeyMaterial::concatenate` logic and improve `Condition<T>` correctness for `i64`/`u64`

## Description

This PR addresses three main issues: a logical bug in `KeyMaterial::concatenate` regarding security strength, the missing implementation of `Condition<u64>`, and a correctness fix for the `Condition<i64>::TRUE` constant.

### 1. Fix func concatenate in KeyMaterial Security Logic

* **Issue:** The documentation for `KeyMaterial::concatenate` states that combining a full entropy key with a low entropy key should result in a **low entropy** key (conserving the weaker security parameters). However, the implementation was using `max()`, which effectively "upgraded" the security strength of the resulting key to the higher of the two inputs.
* **Fix:** Changed the logic to use `min()` for both `key_type` and `security_strength` to strictly adhere to the documentation and safe security modeling principles.

### 2. Implement u64 Condition

* **Issue:** `Condition<u64>` was defined but lacked a concrete implementation, preventing constant-time operations on `u64` primitives.
* **Fix:** Added `impl Condition<u64>` including:
    * Correct mask generation in `from_bool` using `0u64.wrapping_sub(value)`.
    * Correct `TRUE` constant definition as `u64::MAX` (all-ones) to ensure bitwise operations like `select` work correctly across all bits.
    * Implementation of `select`, `is_true`, and other helper methods.

### 3. Fix TRUE Constant in i64 Condition 

* **Issue:** `Condition<i64>::TRUE` was previously defined as `1`. This caused `select` and `negate` functions (which rely on bitwise masking `val & mask`) to fail for all bits except the LSB.
* **Fix:** Updated `Condition<i64>::TRUE` to `-1` (all ones in two's complement) to ensure the mask covers the entire integer width.

### 4. Tests

* Added a new test module `u64_tests` in `ct_tests.rs` to verify `Condition<u64>` behavior.
* Added a specific test case for `or_halves` to verify bit-folding logic.
* Verified that `select` and `negate` logic behaves correctly with the updated `TRUE` constants.

## Motivation

To ensure the library adheres to its stated security contract regarding key material degradation and to provide correct, functional constant-time primitives for 64-bit integers.

## Checklist

- [x] Code compiles correctly.
- [x] Added tests for `Condition<u64>` and `or_halves`.
- [x] `cargo test` passes.
- [x] Documentation comments updated/verified.


ounsworth:
Thank you so much for this!

Unfortunately this is only a mirror so I cannot directly accept pull requests, but I will take a careful look at this for the next alpha, merge changes manually, and include you in a contributors file.

ounsworth:
This has been merged, and will be part of the 0.1.2alpha release.
https://github.com/bcgit/bc-rust/commit/8022962999938b61ccd55dfe1246155d7a69ed42

PR #11, opened 2026-05-27
Fix coding style and remove accidental print

npajkovsky:


ounsworth:
Thanks Nikola, but this has conflicts from the work I merged from branch /feature/mlkem, and my IDE is now using rustfmt on save. So I think I'm going to close this without merging and we can make formatting consistent in a future round of changes.

Issue #6, opened 2026-02-12
Discrepancy between doc and implementation in `KeyMaterial::concatenate`


Quant-TheodoreFelix:
### Title

Discrepancy between documentation and implementation in `KeyMaterial::concatenate`

### Description

I have noticed a contradiction between the rustdoc comments and the actual implementation of the `KeyMaterial::concatenate` method in `crypto/core-interface/src/key_material.rs`.

Documentation states- The documentation explicitly states that concatenating a full entropy key with a low entropy key should downgrade the result to a low entropy key:

> "The resulting [KeyType] and security strength will be the **lesser of the two keys**."
> ...
> "Concatenating a full entropy key with a low entropy key will **result in a low entropy key**."

(Source: `crypto/core-interface/src/key_material.rs` lines 156-164)

Implementation does- However, the implementation uses `max()` instead of `min()`:

```rust
self.key_type = max(&self.key_type, &other.key_type()).clone();
self.security_strength = max(&self.security_strength, &other.security_strength()).clone();
```

(Source: `crypto/core-interface/src/key_material.rs` lines 569-570)

The Conflict- According to the `PartialOrd` implementation for `KeyType`, `BytesLowEntropy` is less than `BytesFullEntropy` (`BytesLowEntropy < BytesFullEntropy`).

Therefore, calling `max()` on these two types will result in `BytesFullEntropy`, effectively "upgrading" the security level of the low entropy key, which contradicts the "safe usage" design philosophy described in the documentation.

Fix- The implementation should likely be changed to use `min()` to adhere to the conservative security model described in the docs.

```rust
// Proposed change
use bouncycastle_utils::min;

// ... inside concatenate ...
self.key_type = min(&self.key_type, &other.key_type()).clone();
self.security_strength = min(&self.security_strength, &other.security_strength()).clone();
```

I hope this issue serves a good purpose...

ounsworth:
Thank you so much for this report!
I am pleased that people are looking at bc-rust in this much detail! ❤️ 

Now that I am thinking about it; I am not sure exactly what the correct behaviour should be.

There is an interaction here between the KeyType and the SecurityStrength. I'm just thinking out loud here: if you concatenate these two values, what would you expect to get?

```
KeyMaterial = {key_len: 32, KeyType: BytesFullEntropy, SecurityStrength: _128bit}
KeyMaterial = {key_len: 32, KeyType: Zeroized, SecurityStrength: none}
```
[...]

PR #12, opened 2026-05-29
addressed doc discrepancy, relates to github #6

dghgit:
Fix (hopefully) for issue #6


ounsworth:
This PR currently causes ml-kem tests to fail. Needs some fixing.

Quant-TheodoreFelix:
It looks like the issue is occurring because `bc_test_data.rs` is referencing an unconnected external repository (`../../../bc-test-data`). The same failure also reproduces on `main`.

ounsworth:
heh. Right. That's a "It works on my computer" problem. Oops.

PR #13, opened 2026-06-03
Proposals for traits for symmetric ciphers

ounsworth:


crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
+use crate::{DUMMY_SEED_512, DUMMY_SEED_1024};
```
officialfrancismendoza, line 1: Currently unused. Maybe you intended to add something later?

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+
+        let security_strengths = [
+            SecurityStrength::None,
+            SecurityStrength::_112bit,
+            SecurityStrength::_128bit,
+            SecurityStrength::_192bit,
+            SecurityStrength::_256bit,
+        ];
+        for ss in security_strengths.iter() {
+            key.set_security_strength(ss.clone()).unwrap();
```
officialfrancismendoza, line 83: Confused: for security strength tests, the loop mutates key, but calls &mac_key? Failure here may be due to wrong key type rather than weak strength as intending.

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+
+            match C::encrypt_out(&mac_key, msg, &mut ct) {
```
officialfrancismendoza, line 322: For this test, you're checking security strength but I think you meant to use key and not mac_key for this for the too strong case?

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+        // todo -- add tests for aead_encrypt() / aead_decrypt() wrapped in a #[cfg(std)]
+
+        // messing with the ciphertext does not give back the same plaintext (or failing to decrypt is also ok)
+        ct[17] ^= 0xFF;
+        match C::aead_decrypt_out(&key, &nonce, aad, &ct[..ct_bytes_written], &tag, &mut pt) {
+            Ok(bytes_written) => {
+                // so it decrypted something, but it had better not match the original plaintext
+                assert_eq!(bytes_written, pt_bytes_written);
+                assert_ne!(&pt[..bytes_written], msg);
+            }
```
officialfrancismendoza, line 257: For AEAD, shouldn't modified ciphertext not decrypt successfully? Instead of silent corrupted plaintext, what about TagCheckFailed or equivalent?

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+            Err(AEADError::TagCheckFailed) => { /* good */ }
+            _ => panic!("Expected TagCheckFailed error"),
+        };
+
+        // multiple invocations give different nonces
+        let (nonce1, _ct_bytes_written, _tag) =
+            C::aead_encrypt_out(&key, aad, msg, &mut ct).unwrap();
+        let (nonce2, _ct_bytes_written, _tag) =
+            C::aead_encrypt_out(&key, aad, msg, &mut ct).unwrap();
+        assert_ne!(nonce1, nonce2);
```
officialfrancismendoza, line 293: Wouldn't this fail for nonce.len() != 0l which was declared earlier?

crypto/core/src/traits.rs
```diff
@@ -17,7 +20,207 @@ pub trait Algorithm {
...
+    /// typically the ciphertext is the same length as the plaintext, but some ciphers may have an expansion factor or require
+    /// extra space for a nonce or tag.
+    /// Returns a tuple containing the initialization data and the number of bytes written to the ciphertext buffer.
+    fn encrypt_out(
+        key: &KeyMaterial<KEY_LEN>,
+        plaintext: &[u8],
+        ciphertext: &mut [u8],
+    ) -> Result<([u8; INIT_DATA_LEN], usize), SymmetricCipherError>;
+    #[cfg(std)]
+    /// A one-shot API to decrypt some ciphertext with the given key.
```
officialfrancismendoza, line 50: Why does decrypt have &self and no init data? It's inconsistent with encrypt, which is static. If we want a one-shot decrypt, shouldn't we also have decrypt as a static?

crypto/core/src/traits.rs
```diff
@@ -17,7 +20,207 @@ pub trait Algorithm {
...
+        plaintext: &mut [u8; BLOCK_LEN],
+    ) -> Result<usize, SymmetricCipherError>;
+}
+
+/// The basic functions of an Authenticated Encryption with Addititional Data cipher.
+pub trait AEADCipher<const KEY_LEN: usize, const NONCE_LEN: usize, const TAG_LEN: usize>:
+    SymmetricCipher<KEY_LEN, NONCE_LEN> + Sized
+{
+    #[cfg(std)]
+    /// A one-shot API to encrypt some plaintext with the given key.
```
officialfrancismendoza, line 121: For a secure one-shot API, why are we taking in a manual nonce? Shouldn't it be automatic? Or at least manual nonce generation API separated/marked hazardous?

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+
+            assert_eq!(msg_chunk, &pt);
+        }
+
+        // test that the iv is random (ie not the same on two runs)
+        let (_encryptor, iv1) = C::do_encrypt_init(&key).unwrap();
+        let (_encryptor, iv2) = C::do_encrypt_init(&key).unwrap();
+        assert_ne!(iv1, iv2);
+
+        // error case: KeyMaterial of wrong type
```
officialfrancismendoza, line 159: Just so I am understanding correctly: to trigger the wrong keyMaterial error in these sort of cases, don't we want to pass in mac_key instead of normal key?

See this pattern in your other wrong keyMaterial test cases, where you have initialized one key but pass in the other type into do_encrypt_init, encrypt_out, or equivalent metrics. Want to deconflcit which is the (technically wrong) key we want to pass in to do these error checks.

officialfrancismendoza (changes requested):
Wanted to check possible bugs in tests and the AEAD tampering test potentially allowing successful corrupted decrypts.

officialfrancismendoza:
@ounsworth merge conflict in QUALITY_AND_STYLE.md. Also had a few unresolved comments about presumably mismatched key usage (ie: mac_key instead of key and vice versa), permuting over a different key, etc.

crypto/core/src/traits.rs
```diff
@@ -17,7 +19,207 @@ pub trait Algorithm {
     const MAX_SECURITY_STRENGTH: SecurityStrength;
 }
 
-pub trait Hash : Default {
+/// The basic one-shat encrypt and decrypt that all types of symmetric ciphers must implement.
```
romen: Funny typo ? Or intentionally sassy?

officialfrancismendoza: lol

crypto/core-test-framework/src/symmetric_ciphers.rs
ounsworth: Yep, that's a bug. Well spotted.

crypto/core-test-framework/src/symmetric_ciphers.rs
ounsworth: I don't understand.

crypto/core/src/traits.rs
ounsworth: Yep, that's right.

ounsworth:
> Wanted to check possible bugs in tests and the AEAD tampering test potentially allowing successful corrupted decrypts.

Well-spotted. I'm I have you as a second pair of eyes! As I said, I wrote these traits and tests without any actual code, so I entrust this to you: feel free to adjust any of it as you develop ascon.

crypto/core/src/traits.rs
ounsworth: OOOPS.

crypto/core/src/traits.rs
officialfrancismendoza: ⚠️**Important Design Pattern**: Since we are catering for two different strata of developers (novices vs experts) , base trait must _never_ expose invasive functionality where uninitiated can make catastrophic errors (ie: nonce reuse, no seed/poor seed for randomness, etc.). 

But there are a set of external methods to allow _experienced_ developers to make more invasive said changes given their needs (ie: source their own RNG instead of using ours, etc.)

For traits, supposed to be generic enough to apply to multiple cryptosystems, but impl-ing for more specific traits (ie: aeadCipher should not overindex on ASCON, etc.)

officialfrancismendoza:
LGTM!

PR #16, opened 2026-06-08
Tackling the 0.1.2alpha TODO list

ounsworth:
I am going to collect here a bunch of small changes as I work through the TODO list, and then can do a squash merge at the end.

PR #17, opened 2026-06-09
Zeroize output buffers at the start of all _out functions (part of #16)

Quant-TheodoreFelix:
Summary

This zeroizes the caller-provided output buffer at the entry of every concrete function that writes into an output byte slice (the `*_out` family, the dispatch wrappers in the `factory` crate, and the constant-time copy helper in `utils`). Each such function now starts with `out.fill(0)` (or `output.fill(0)`) before it produces its result. 84 write sites in total.

Motivation (from the 0.1.2alpha TODO)

The TODO states: anywhere you have an `_out(.. out: &mut [u8])`, start by zeroizing it with `.fill(0)`. Pre-zeroizing closes two leak paths:
- An oversized output buffer keeps stale bytes past the digest/MAC length.
- An early error return leaves partially written or stale data in the caller's buffer.

A representative change:

```rust
fn do_final_out(mut self, output: &mut [u8]) -> usize {
    output.fill(0);
    ...
}
```

Scope

`sha2`, `sha3` (`sha3` / `shake` / `keccak`), `hmac`, `rng` (`hash_drbg`), `hex`, `base64`, `factory` (`hash` / `xof` / `mac` / `rng`), `utils/ct`, and the PQ crates `mlkem`, `mlkem_lowmemory`, `mldsa`, `mldsa_lowmemory` (key-encode and sign paths).

Behavior change (please note)

Digest, MAC, XOF and RNG output functions now zero the tail of an oversized output buffer instead of leaving it untouched. One existing `sha3` unit test (`test_static_hash`) asserted the old "does not touch the extra bytes" contract and was updated to assert the new zeroed-tail contract. The shared `core-test-framework` hash and mac tests already use zero-initialized buffers and expect zeroed tails, so they were unaffected.

Deliberately excluded

The row-wise partial writers in `mlkem_lowmemory` (`pack_t_hat_row`, `pack_s_hat_row`, `compress_u_row`) were NOT changed. They write a sub-slice of a shared buffer that already holds other rows, so a blanket `fill(0)` would erase previously written data and corrupt the output. Trait declarations (no body), `unimplemented!()` stubs, and `KeyMaterialTrait`-typed outputs (zeroized transitively through the byte-slice writers) were left as-is.

Merge conflict warning

This PR modifies `crypto/rng/src/hash_drbg80090a.rs` (`generate_out`, `next_bytes_out`, `hash_df`, `hashgen`). PR #16 also touches that file, so expect a conflict when folding this into the roundup branch.

Verification

```
cargo build            # clean across the workspace
cargo test --workspace # 326 passed, 0 failed
```

The test run includes the ML-KEM / ML-DSA wycheproof KATs, which exercise the modified key-encode and sign paths.

Relationship

Part of #16. Intended to be folded into the 0.1.2alpha TODO roundup.

### DIFF
--- a/crypto/base64/src/lib.rs
+++ b/crypto/base64/src/lib.rs
@@ -193,6 +193,8 @@ impl Base64Encoder {
         assert!(inref.len() >= 3);
         assert!(out.len() >= 4);
 
+        out.fill(0);
+
         out[0] = Self::ct_bin_to_b64(inref[0] >> 2);
         out[1] = Self::ct_bin_to_b64(((inref[0] & 0x03) << 4) | inref[1] >> 4);
         out[2] = Self::ct_bin_to_b64(((inref[1] & 0x0F) << 2) | inref[2] >> 6);
--- a/crypto/factory/src/hash_factory.rs
+++ b/crypto/factory/src/hash_factory.rs
@@ -129,6 +129,8 @@ impl Hash for HashFactory {
     }
 
     fn hash_out(self, data: &[u8], output: &mut [u8]) -> usize {
+        output.fill(0);
+
         match self {
             Self::SHA224(h) => h.hash_out(data, output),
             Self::SHA256(h) => h.hash_out(data, output),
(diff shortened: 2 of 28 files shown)


Quant-TheodoreFelix:
Superseded by #18 (head branch renamed from qtf/0.1.2alpha-zeroize-out-buffers to qtfelix/0.1.2alpha-zeroize-out-buffers; the rename auto-closed this cross-fork PR). Same commit and contents.

PR #18, opened 2026-06-09
Zeroize output buffers at the start of all _out functions (part of #16)

Quant-TheodoreFelix:
This zeroizes the caller-provided output buffer at the entry of every concrete function that writes into an output byte slice (the `*_out` family, the dispatch wrappers in the `factory` crate, and the constant-time copy helper in `utils`). Each such function now starts with `out.fill(0)` (or `output.fill(0)`) before it produces its result. 84 write sites in total.

Motivation (from the 0.1.2alpha TODO)

The TODO states: anywhere you have an `_out(.. out: &mut [u8])`, start by zeroizing it with `.fill(0)`. Pre-zeroizing closes two leak paths:
- An oversized output buffer keeps stale bytes past the digest/MAC length.
- An early error return leaves partially written or stale data in the caller's buffer.

A representative change:

```rust
fn do_final_out(mut self, output: &mut [u8]) -> usize {
    output.fill(0);
    ...
}
```

Scope

`sha2`, `sha3` (`sha3` / `shake` / `keccak`), `hmac`, `rng` (`hash_drbg`), `hex`, `base64`, `factory` (`hash` / `xof` / `mac` / `rng`), `utils/ct`, and the PQ crates `mlkem`, `mlkem_lowmemory`, `mldsa`, `mldsa_lowmemory` (key-encode and sign paths).

Behavior change (please note)

Digest, MAC, XOF and RNG output functions now zero the tail of an oversized output buffer instead of leaving it untouched. One existing `sha3` unit test (`test_static_hash`) asserted the old "does not touch the extra bytes" contract and was updated to assert the new zeroed-tail contract. The shared `core-test-framework` hash and mac tests already use zero-initialized buffers and expect zeroed tails, so they were unaffected.

Deliberately excluded

The row-wise partial writers in `mlkem_lowmemory` (`pack_t_hat_row`, `pack_s_hat_row`, `compress_u_row`) were NOT changed. They write a sub-slice of a shared buffer that already holds other rows, so a blanket `fill(0)` would erase previously written data and corrupt the output. Trait declarations (no body), `unimplemented!()` stubs, and `KeyMaterialTrait`-typed outputs (zeroized transitively through the byte-slice writers) were left as-is.

Merge conflict warning

This PR modifies `crypto/rng/src/hash_drbg80090a.rs` (`generate_out`, `next_bytes_out`, `hash_df`, `hashgen`). PR #16 also touches that file, so expect a conflict when folding this into the roundup branch.

Verification

```
cargo build            # clean across the workspace
cargo test --workspace # 326 passed, 0 failed
```

The test run includes the ML-KEM / ML-DSA wycheproof KATs, which exercise the modified key-encode and sign paths.

Relationship

Part of #16. Intended to be folded into the 0.1.2alpha TODO roundup.

### DIFF
--- a/crypto/mlkem/src/matrix.rs
+++ b/crypto/mlkem/src/matrix.rs
@@ -171,8 +171,6 @@ impl<const k: usize> Vector<k>
         // let mut s = self.clone();
         // s.conditional_sub_q();
 
-        out.fill(0);
-
         let mut idx = 0;
         match du {
             10 => { // MLKEM512 and MLKEM 768
--- a/crypto/mlkem/src/matrix.rs
+++ b/crypto/mlkem/src/matrix.rs
@@ -171,6 +171,8 @@ impl<const k: usize> Vector<k>
         // let mut s = self.clone();
         // s.conditional_sub_q();
 
+        out.fill(0);
+
         let mut idx = 0;
         match du {
             10 => { // MLKEM512 and MLKEM 768
--- a/crypto/mlkem/src/polynomial.rs
+++ b/crypto/mlkem/src/polynomial.rs
@@ -127,8 +127,6 @@ impl Polynomial {
         // each of the N i16's will take dv bits
         debug_assert_eq!(out.len(), N * (dv as usize) / 8);
 
-        out.fill(0);
-
         let mut t = [0u8; 8];
         let mut idx = 0;
 
--- a/crypto/mlkem/src/polynomial.rs
+++ b/crypto/mlkem/src/polynomial.rs
@@ -127,6 +127,8 @@ impl Polynomial {
         // each of the N i16's will take dv bits
         debug_assert_eq!(out.len(), N * (dv as usize) / 8);
 
+        out.fill(0);
+
         let mut t = [0u8; 8];
         let mut idx = 0;
 
--- a/crypto/mlkem_lowmemory/src/polynomial.rs
+++ b/crypto/mlkem_lowmemory/src/polynomial.rs
@@ -150,8 +150,6 @@ impl Polynomial {
         // each of the N i16's will take dv bits
         debug_assert_eq!(out.len(), N * (dv as usize) / 8);
 
-        out.fill(0);
-
         let mut t = [0u8; 8];
         let mut idx = 0;
 
--- a/crypto/mlkem_lowmemory/src/polynomial.rs
+++ b/crypto/mlkem_lowmemory/src/polynomial.rs
@@ -150,6 +150,8 @@ impl Polynomial {
         // each of the N i16's will take dv bits
         debug_assert_eq!(out.len(), N * (dv as usize) / 8);
 
+        out.fill(0);
+
         let mut t = [0u8; 8];
         let mut idx = 0;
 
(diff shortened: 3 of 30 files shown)


crypto/mlkem/src/matrix.rs
```diff
@@ -171,6 +171,8 @@ impl<const k: usize> Vector<k>
         // let mut s = self.clone();
         // s.conditional_sub_q();
 
+        out.fill(0);
```
ounsworth: This one should be removed -- this is not a public function, and this will have a performance impact on the mlkem algorithm.

crypto/mlkem/src/polynomial.rs
```diff
@@ -127,6 +127,8 @@ impl Polynomial {
         // each of the N i16's will take dv bits
         debug_assert_eq!(out.len(), N * (dv as usize) / 8);
 
+        out.fill(0);
```
ounsworth: This one should be removed -- this is not a public function, and this will have a performance impact on the mlkem algorithm.

crypto/mlkem_lowmemory/src/polynomial.rs
```diff
@@ -150,6 +150,8 @@ impl Polynomial {
         // each of the N i16's will take dv bits
         debug_assert_eq!(out.len(), N * (dv as usize) / 8);
 
+        out.fill(0);
```
ounsworth: This one should be removed -- this is not a public function, and this will have a performance impact on the mlkem algorithm.

crypto/sha3/tests/sha3_tests.rs
```diff
@@ -66,10 +66,13 @@ mod sha3_tests {
         assert_eq!(&out, b"\x58\x4c\xc7\x02\xc2\x22\x9a\x0a\xbc\x78\x9b\xfa\x64\xb4\x27\x1f\xb8\xf0\xbb\x78\x67\x15\x88\xb9\xef\x1d\x09\x3e\xa3\xd4\x72\x58\x4c\x6d\x43\xb5\x68\x33\x59\x47\x2f\x44\x1b\x33\x85\x6f\x68\x28\x59\xf0\xc3\x95\x4b\x56\x80\x8f\xd1\xfb\xa0\xb5\x9c\x9d\x19\x54");
         assert_eq!(bytes_written, 64);
 
-        // check that if you feed it an output slice that's bigger than it needs, that it doesn't touch the extra bytes.
+        // Q. T. Felix NOTE: With the application of zeroize, the entire output buffer is pre-initialized to 0, 
+        //                   so the bytes after the digest length are now also 0. 
+        //                   Previously, the contract was to leave the trailing bytes untouched, 
+        //                   but this has been changed to fill them with zeros to prevent exposure of stale data.
```
ounsworth: A comment to explain what is being tested by this line is good, but can you please remove your name, and can you please remove the description of what changed from one version to the next? (the change history will very quickly become meaningless)

crypto/utils/src/ct.rs
```diff
@@ -292,6 +292,8 @@ pub fn conditional_copy_bytes<const LEN: usize>(
     out: &mut [u8; LEN],
     take_a: bool,
 ) {
+    out.fill(0);
```
ounsworth: This needs to be removed as it may violate the constant-time assumption since over-writing a zero with a zero may take less time than overwriting a different value.

ounsworth (changes requested):
This is great, thanks!

I have left a few requests for changes.

Can you please also update the docstring in core/traits.rs for each *_out function to indicate that the entire output buffer will be zeroized.

This PR is marked Draft. Please let me know when you think it is ready to be merged.

If you have done a task from the todo list, can you please delete that line and move it to the changelog section at the bottom of that file?

Quant-TheodoreFelix:
Yes, confirmed. Im working on it.

Quant-TheodoreFelix:
Thanks for the review! All requested changes are in (commit c642ef6):

- Removed the `fill(0)` from the three internal mlkem compress functions (perf) and from `ct::conditional_copy_bytes` (constant-time concern).
- Cleaned up the sha3 test comment (no name, no change history).
- Added "the entire output buffer is zeroized" to the docstrings of the `*_out` functions in core/traits.rs.
  One deliberate exception: `derive_key_out`, `derive_key_from_multiple_out` and `fill_keymaterial_out` take a `KeyMaterialTrait` output rather than a byte slice, and HKDF only writes the first L bytes of the key's internal buffer, so a blanket zeroize claim there would be inaccurate. Happy to extend it if you prefer.
- Moved the TODO line to the changelog section of the release notes.

`cargo test --workspace` passes (326 tests). Marked ready for review.

ounsworth (approved):

ounsworth:
Amazing! Thank you so much! I will merge now.

PR #21, opened 2026-06-16
Feat: Implement ASCON Lightweight Cryptographic Libraries 

officialfrancismendoza:
## Description
Implementation of NIST SP 800-232 ASCON lightweight cryptography standards for constrained devices (#22), including:

- Authenticated Encryption (aead128)
- Hashing (hash256)
- Extendable Output Functions (xof128)
- Custom Extendable Output Functions (cxof128)

**Note**: due to the important considerations of embedded systems, this PR must follow, where applicable, utilization of non-heap memory structures (for example: [u8, CONST_SIZE] instead of vec<u8>) to respect their compute and RAM limitations.

## Type of Change
- Changes to core crate, affecting traits
- A new ASCON crate with associated src files and testing suite

## Testing
- `cargo test -p bouncycastle-ascon` to test ASCON crate and associated KAT test cases
- `cargo test --workspace`
- `cargo mutants -p bouncycastle-ascon` for mutant tests

Verification results (so reviewers don't have to re-derive them):
- NIST LWC KAT sweeps against `bc-test-data` (one-shot and byte-at-a-time streaming): 1089 Ascon-AEAD128, 1025 Ascon-Hash256, 1025 Ascon-XOF128, 1089 Ascon-CXOF128 -- 4228 vectors total, all passing.
- `cargo mutants -p bouncycastle-ascon`: 616 mutants, 537 caught, 4 missed, 75 unviable, 0 timeouts. All 4 missed mutants are equivalent mutants (two branch conditions that degenerate to a no-op on the boundary value, and a disjoint-bitmask `|`/`^` equivalence in `set_state_byte`), not real gaps.

The `get_test_data` loader in `crypto/ascon/tests/bc_test_data.rs` is a copy of the same helper already used by the mlkem/mldsa crates; consolidating the duplicated copies into one shared helper is deferred to a separate PR.

Wycheproof coverage is pending upstream SP 800-232 test vectors: the existing Wycheproof Ascon vectors target the pre-standard CAESAR variant, not SP 800-232, so no Wycheproof harness is included here. Crucible does not yet have Ascon vectors either.

### DIFF
--- a/crypto/ascon/src/ascon_aead128.rs
+++ b/crypto/ascon/src/ascon_aead128.rs
@@ -1,0 +1,724 @@
+//! Ascon-AEAD128 authenticated encryption, as specified in NIST SP 800-232 §4.
+//!
+//! Rate = 128 bits, capacity = 192 bits, 128-bit key/nonce/tag. Initialization and finalization use
+//! `Ascon-p[12]`; associated-data and plaintext/ciphertext blocks use `Ascon-p[8]`.
+//!
+//! Every byte of plaintext/ciphertext is transformed and emitted as soon as it is seen (no
+//! held-back buffering across `do_encrypt_update`/`do_decrypt_update` calls); this is what lets the
+//! finalizers be plain `self -> tag` / `self -> Result<(), _>` calls with nothing left to flush.
+//! Ascon-AEAD128 permits this because within a 128-bit rate block each plaintext/ciphertext byte
+//! is transformed independently of the others in that block; the permutation only runs once a
+//! full 16-byte block has been absorbed, or at finalization.
+
+use core::fmt::{self, Debug, Display, Formatter};
+
+use bouncycastle_core::errors::{KeyMaterialError, SuspendableError, SymmetricCipherError};
+use bouncycastle_core::key_material::{KeyMaterial, KeyMaterialTrait, KeyType};
+use bouncycastle_core::suspendable_state::{add_lib_ver, check_lib_ver};
+use bouncycastle_core::traits::{
+    AEADCipher, Algorithm, RNG, SecurityStrength, SuspendableKeyed, SymmetricCipher,
+};
+use bouncycastle_rng::HashDRBG_SHA512;
+use bouncycastle_utils::ct::ct_eq_bytes;
+use bouncycastle_utils::secret::Secret;
+
+use crate::permutation::{AsconState, load_u64_le, p8, p12, store_u64_le};
+
+/// Length in bytes of the Ascon-AEAD128 key.
+pub const KEY_LEN: usize = 16;
+/// Length in bytes of the Ascon-AEAD128 nonce.
+pub const NONCE_LEN: usize = 16;
+/// Length in bytes of the Ascon-AEAD128 authentication tag.
+pub const TAG_LEN: usize = 16;
+const RATE: usize = 16;
+
+/// Ascon-AEAD128 initial value (SP 800-232 Table 14).
+const ASCON_IV: u64 = 0x00001000808C0001;
+
+/// State machine for enforcing the call order and remembering the direction (encrypt/decrypt).
+#[derive(Clone, Copy, PartialEq, Eq, Debug)]
+enum StateMachine {
+    EncInit,
+    EncAad,
+    EncData,
+    DecInit,
+    DecAad,
+    DecData,
+}
+
+impl StateMachine {
+    // Stable u8 encoding used when suspending/resuming the AEAD state machine.
+    fn to_u8(self) -> u8 {
+        match self {
+            StateMachine::EncInit => 0,
+            StateMachine::EncAad => 1,
+            StateMachine::EncData => 2,
+            StateMachine::DecInit => 4,
+            StateMachine::DecAad => 5,
+            StateMachine::DecData => 6,
+        }
+    }
+
+    fn from_u8(v: u8) -> Option<Self> {
+        Some(match v {
+            0 => StateMachine::EncInit,
+            1 => StateMachine::EncAad,
+            2 => StateMachine::EncData,
+            4 => StateMachine::DecInit,
+            5 => StateMachine::DecAad,
+            6 => StateMachine::DecData,
+            _ => return None,
+        })
+    }
+
+    fn is_encrypt(self) -> bool {
+        matches!(self, StateMachine::EncInit | StateMachine::EncAad | StateMachine::EncData)
+    }
+
+    fn is_init(self) -> bool {
+        matches!(self, StateMachine::EncInit | StateMachine::DecInit)
+    }
+}
+
+/// An implementation of the Ascon-AEAD128 algorithm (NIST SP 800-232).
+///
+/// A single instance performs one operation (encryption or decryption) under one (key, nonce) pair.
+/// See [`AsconAead128::new`] for the streaming workflow and [`AsconAead128::encrypt`] /
+/// [`AsconAead128::decrypt`] for the one-shot APIs.
+#[derive(Clone)]
+pub struct AsconAead128 {
+    // 128-bit secret key (two 64-bit words). It is re-added to the state at finalization, so it must
+    // be retained; wrapped in `Secret` for volatile-write zeroization on drop.
+    key: Secret<[u64; 2]>,
+    // 320-bit internal state (five 64-bit words). Carries keystream/plaintext-derived material, so
+    // it is likewise wrapped in `Secret`.
+    state: Secret<AsconState>,
+    // Byte position (0..RATE) within the current rate block.
+    pos: usize,
+    // State machine for enforcing the call order and remembering the direction.
+    state_machine: StateMachine,
+}
+
+impl AsconAead128 {
+    /// Validate a [`KeyMaterial`] for use with Ascon-AEAD128 and return its key words.
+    /// The key must be tagged as a [`KeyType::SymmetricCipherKey`] and carry at least the
+    /// algorithm's 128-bit security strength (SP 800-232 R1/R2).
+    fn checked_key(key: &KeyMaterial<KEY_LEN>) -> Result<[u64; 2], SymmetricCipherError> {
+        if key.key_type() != KeyType::SymmetricCipherKey {
+            return Err(KeyMaterialError::InvalidKeyType(
+                "Ascon-AEAD128 requires a SymmetricCipherKey",
+            )
+            .into());
+        }
+        if key.security_strength() < SecurityStrength::_128bit {
+            return Err(KeyMaterialError::SecurityStrength(
+                "Ascon-AEAD128 requires a key with at least 128-bit security strength",
+            )
+            .into());
+        }
+        let bytes = key.ref_to_bytes();
+        if bytes.len() != KEY_LEN {
+            return Err(KeyMaterialError::InvalidLength.into());
+        }
+        Ok([load_u64_le(bytes, 0), load_u64_le(bytes, 8)])
+    }
+
+    /// Draw a fresh, unique 128-bit nonce from the library's default OS-seeded DRBG.
+    ///
+    /// The one-shot APIs of main's cipher framework generate the init data / nonce internally, so
+    /// Ascon's per-encryption nonce-uniqueness requirement (SP 800-232 R3) is satisfied by sourcing
+    /// each nonce from a CSPRNG. Callers who need deterministic, caller-supplied nonces should use
+    /// the inherent streaming API ([`AsconAead128::new`]).
+    fn fresh_nonce() -> Result<[u8; NONCE_LEN], SymmetricCipherError> {
+        let mut rng = HashDRBG_SHA512::new_from_os();
+        let mut nonce = [0u8; NONCE_LEN];
+        rng.next_bytes_out(&mut nonce)?;
+        Ok(nonce)
+    }
+
+    /// Create a new streaming instance.
+    /// * `key` is validated as a [`KeyType::SymmetricCipherKey`] with at least 128-bit strength.
+    /// * `nonce` is the 128-bit nonce. It **must** be unique per encryption under a given key.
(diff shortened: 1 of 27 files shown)


ounsworth:
@officialfrancismendoza I see this. On my todo list!

jjkurczak:
@officialfrancismendoza, a couple things on this one so far:

1. cargo mutants is seeing 124 missed mutations - it's hard for me to judge how many of them are legitimate, but is this similar to the volume you were seeing yourself?
2. I'm seeing your .claude/settings.json being pushed into the repo, might be worth using https://git-scm.com/docs/gitignore to keep this out of commits automatically?

ounsworth (commented):
First review:

This PR is currently failing the rustfmt check. Just run `$ cargo fmt` in the root and commit again.

[The Crucible project](https://github.com/symbolicsoft/crucible) only has tests for ML-KEM and ML-DSA. So does not apply here.

There are ascon vectors in both [Wycheproof](https://github.com/C2SP/wycheproof) and [bc-test-data](https://github.com/bcgit/bc-test-data), so you should be able to take the test harnesses that we have in, for example, ml-dsa and port that over for ascon.

(I will continue with a code review, though I admit that +3,900 is a lot to review, so I will have to do this in chunks.)

crypto/ascon/summary.md
```diff
@@ -0,0 +1,262 @@
+# ASCON Implementation & Testing — Work Summary
```
ounsworth: This file seems maybe like it is your personal notes that were checked in by accident? I don't see any long-term value in including this file in the source code.

crypto/ascon/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+//!
+//! - **Nonce uniqueness (SP 800-232 R3):** a (key, nonce) pair must never be reused for two
+//!   different Ascon-AEAD128 encryptions. Nonce reuse breaks confidentiality.
+//! - **Tag length:** this crate always produces and verifies the full 128-bit tag. Truncated tags
+//!   (SP 800-232 §4.2.1) are not exposed.
+//! - **Key handling:** sensitive fields (the key and the working permutation state) are held in
+//!   [`bouncycastle_utils::secret::Secret`] wrappers, so they are scrubbed with volatile writes
+//!   when the value is dropped. The hash/XOF states are likewise `Secret`-wrapped.
+//! - **Decryption release:** never release decrypted plaintext until finalization returns `Ok`; an
+//!   `Err(AuthenticationFailed)` means the ciphertext or tag was tampered with.
```
ounsworth: I like this point and I think it should be expanded. I'm thinking something like: 

> **Decryption tag check failure**: a ciphertext decryption whose finalization returns an `Err(AuthenticationFailed)` should be considered to be tampered with and the entire plaintext should be rejected. For example, if the plaintext being decrypted is large enough that it must be processed by the application in a streaming fashion, the application should have a way to cancel the operation or transaction with an error if ASCON finalization returns an AuthenticationFailed error.

crypto/ascon/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+//! // One-shot:
+//! let digest = AsconHash256::digest(b"hello world");
+//! assert_eq!(digest.len(), 32);
+//!
+//! // Streaming:
+//! let mut h = AsconHash256::new();
+//! h.update_bytes(b"hello ");
+//! h.update_bytes(b"world");
+//! let mut out = [0u8; 32];
+//! h.do_final_into(&mut out);
```
ounsworth, line 27: The rest of the library uses `do_final_out()` for this pattern. I actually kinda like `_into` instead of `_out`, but one way or another this should be made consistent with the rest of the library.

If you think that `_into` is the better name, than why don't you open a github issue to propose that change, and then create a PR against 0.1.3alpha.

crypto/ascon/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+//! assert_eq!(out, digest);
+//! ```
+//!
+//! Authenticated encryption (one-shot):
+//! ```
+//! use bouncycastle_ascon::ascon_aead128::AsconAead128;
+//!
+//! let key = [0u8; 16];
```
ounsworth, line 35: Why is this not a KeyMaterial?

crypto/ascon/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+//! let nonce = [1u8; 16];           // MUST be unique per encryption under a given key
+//! let ad = b"associated data";
+//! let plaintext = b"secret message";
+//!
+//! let mut ct = vec![0u8; plaintext.len() + 16]; // ciphertext || 16-byte tag
+//! let n = AsconAead128::encrypt(&key, &nonce, Some(ad), plaintext, &mut ct);
+//! ct.truncate(n);
```
ounsworth, line 42: This API pattern seems extremely weird to me.
First the user has to create a mut vec of the correct size `plaintext.len() + 16`.
Then the user has to hand that in to the encrypt function.
Then the user has to truncate it because it might actually be shorter than `plaintext.len() + 16` ?

I might expect some acrobatics like that in no_std where the user has to hand in a `[u8; N]` of the correct size, but the whole point of Vec's is to avoid making the user deal with array lengths.
If we're using vec here, then can't we hide this inside the function call and `-> Vec<u8>` that we size correctly?

crypto/ascon/src/lib.rs
```diff
@@ -0,0 +1,100 @@
...
+pub mod ascon_xof128;
+
+/// Algorithm name for Ascon-AEAD128.
+pub const ASCON_AEAD128_NAME: &str = "Ascon-AEAD128";
+/// Algorithm name for Ascon-Hash256.
+pub const ASCON_HASH256_NAME: &str = "Ascon-Hash256";
+/// Algorithm name for Ascon-XOF128.
+pub const ASCON_XOF128_NAME: &str = "Ascon-XOF128";
+/// Algorithm name for Ascon-CXOF128.
+pub const ASCON_CXOF128_NAME: &str = "Ascon-CXOF128";
```
ounsworth, line 100: All the ASCON variants should impl `traits::Algorithm` and maybe `traits::HashAlgParams`?

crypto/ascon/src/util.rs
```diff
@@ -0,0 +1,20 @@
+//! Internal little-endian load/store helpers.
+//!
+//! These replace the external `arrayref` crate so that this crate carries no third-party runtime
+//! dependencies (per the project's QUALITY_AND_STYLE rules). All callers pass slices that are at
+//! least 8 bytes long at the given offset, so `copy_from_slice` is infallible by construction and
+//! no fallible conversion is involved.
```
ounsworth, line 6: Claude likes to leave comments like these explaining its reasoning. That's fine for you to understand what it's done, but should be deleted before committing.

crypto/ascon/src/ascon_aead128.rs
```diff
@@ -0,0 +1,940 @@
...
+        self.s[3] = t3 ^ t3.rotate_right(10) ^ t3.rotate_right(17);
+        self.s[4] = t4 ^ t4.rotate_right(7) ^ t4.rotate_right(41);
+    }
+
+    /// Returns a 64-bit value with a single "1" at bit position (i*8): the integer form of the
+    /// padding rule for a partial block (SP 800-232 Appendix A.2).
+    fn pad(i: usize) -> u64 {
+        debug_assert!(i < 8);
+        0x01u64 << (i * 8)
+    }
```
ounsworth, line 232: I assume that this function should correspond to Algorithm 2 in SP.800-232.
I would like you to use a commenting style similar to what I have done in ML-DSA to make it clear how our code corresponds with the sample code in SP.800-232.
For example:
https://github.com/bcgit/bc-rust/blob/main/crypto/mldsa/src/aux_functions.rs#L15

Moreover, for larger functions, I actually comment line-by-line with the corresponding line numbers in the FIPS sample algorithms to make the correspondence super clear, especially if our code is skipping steps or doing steps out-of-order.
Some examples:
https://github.com/bcgit/bc-rust/blob/main/crypto/mldsa/src/aux_functions.rs#L435
https://github.com/bcgit/bc-rust/blob/main/crypto/mldsa/src/mldsa.rs#L929

This also provides a way to note any deviations that we are making from the FIPS, which will make certification easier.
Example:
https://github.com/bcgit/bc-rust/blob/main/crypto/mldsa/src/mldsa_keys.rs#L665

But actually, looking closer, this does not appear to be the same as Algorithm 2, this is generating the padding string `1 || 0^j` but not performing the concatenation. So then it is confusing to give it the same name as Alg 2.

Possibly, this indicates that in fact this code does not correspond directly to the FIPS sample algorithms and we should do a refactor so that it does?

crypto/ascon/src/ascon_xof128.rs
```diff
@@ -0,0 +1,307 @@
...
+    }
+
+    // Pad the final absorbed block (SP 800-232 §5.2 step 3 / Alg. 6).
+    fn pad_and_absorb(&mut self) {
+        let final_bits = (self.buf_pos << 3) as u32;
+        let mask: u64 = if final_bits == 0 {
+            0x00FFFFFFFFFFFFFF
+        } else {
+            0x00FFFFFFFFFFFFFF >> (56 - final_bits)
+        };
```
ounsworth, line 135: This function exists verbatim in both `ascon_xof128.rs` and `ascon_cxof128.rs`. That means that some de-duplication should be done.

I would suggest doing some re-structuring of this code to better match the structure suggested by SP.800-232:
* Have an `aux_functions.rs` that corresponds to SP.800-232 section 2.1 which includes Algorithm 1 parse(X, r) and Algorithm 2 pad(X, r), and then also serves as a logical place to put other code that will be common to the various parts of ACSON.
* Maybe create `permutation.rs` that corresponds with SP.800-232 and contains the implementations of 𝐴𝑠𝑐𝑜𝑛-𝑝[8] and 𝐴𝑠𝑐𝑜𝑛-𝑝[12]. I would expect to see this as a single parametrized function to avoid duplicate code, reduce code-review burden, and reduce number of tests needed.

crypto/ascon/src/ascon_aead128.rs
```diff
@@ -0,0 +1,940 @@
...
+                needed,
+            ));
+        }
+        let nonce = Self::fresh_nonce()?;
+        // No associated data for the plain SymmetricCipher view; the tag is appended to `ciphertext`.
+        let mut cipher = Self::new(&k, &nonce, None, true);
+        let n = cipher.encrypt_update(plaintext, ciphertext);
+        let written = cipher.encrypt_finalize(&mut ciphertext[n..]);
+        Ok((nonce, n + written))
+    }
```
ounsworth, line 695: I assume this `encrypt_out` function is supposed to correspond to Algorithm 3 Ascon-AEAD128.enc(𝐾, 𝑁 , 𝐴, 𝑃 ) from SP.800-232?

A few things that catch my eye on a first look:
* This should match exactly the API of Algorithm 3, including the name `enc` `enc_out` instead of `encrypt` / `encrypt_out`. I would also make the names of the inputs `K, N, A, P` match exactly. 
* A header comment in the same style as, for example, [this](https://github.com/bcgit/bc-rust/blob/main/crypto/mldsa/src/mldsa.rs#L1120) indicating that this is the implementation of Algorithm 3.
* Algorithm 3 has 4 inputs, but this only has 3. I see that you are not allowing the user to provide the nonce, but instead you are forcing them to use our `HashDRBG_SHA512::new_from_os()` (via the `Self::fresh_nonce()` call). This will be problematic, for example, in environments that can't use our RNG and need to connect in for example a hardward RNG. I think we should match exactly the API of FIPS Algorithm 3. If you want, you could do
`N: Option<&[u8; NONCE_LEN]>` (or maybe that should be a KeyMaterial?) and if it's None then we can generate it for the user. I've done similar patterns in ML-DSA / ML-KEM.
You should also provide an `enc_rng` similar to what I've done with [MLKEM.encaps_rng](https://github.com/bcgit/bc-rust/blob/main/crypto/mlkem/src/mlkem.rs#L943) that takes an ` &mut dyn RNG` and uses that for sourcing entropy.
* The name of the constant `CRYPTO_KEYBYTES` is confusing. I think this should be `KEY_LEN` to be consistent with other similar constants across the library (I use "LEN" for a length in bytes, and "SIZE" for lengths in bits. This is described in the [QUALITY_AND_STYLE.md](https://github.com/bcgit/bc-rust/blob/main/QUALITY_AND_STYLE.md#naming-conventions). Same goes for `CRYPTO_ABYTES`.
* At a quick glance, the body of this function does not look anything like the the body of Algorithm 3; I think because you have chopped up Algorithm 3 into a new(), update(), final() streaming pattern, which is fine, but you'll need to heavily comment all of the involved functions to make the correspondence with the FIPS Algorithms clear.
* By the way, the rest of the library uses `_final()` not `_finalize()`.
* Algorithm 3 returns Output: ciphertext 𝐶, 128-bit tag 𝑇. You are returning the nonce, and it doesn't look like you're returning the tag at all. This doesn't make sense to me.

officialfrancismendoza:
@dghgit remediated concerns based on plan and changes to release/0.1.3alpha

cpu:
> Wycheproof coverage is pending upstream SP 800-232 test vectors: the existing Wycheproof Ascon vectors target the pre-standard CAESAR variant, not SP 800-232, so no Wycheproof harness is included here.

FWIW this was resolved upstream w/ https://github.com/C2SP/wycheproof/pull/267

officialfrancismendoza:
The most recent changes for ASCON, which utilize the core API change for splitting into a AEADCipherEncryptor and AEADCipherDecryptor (among other changes), has been included as part of the AEAD cipher split API (#120) off of feature branch (#118) and then subsequently merged into [bcgit:feature/xof-cshake](https://github.com/bcgit/bc-rust/tree/feature/xof-cshake). Hence, closing this PR as it is stale.

Issue #32, opened 2026-06-18
Apply Secret trait consistently across the library


ounsworth:
Some of the earlier crates (SHA2 / SHA3) were done before I added the Secret trait, so circle back and add it, and impl all the traits that it requires.

At the same time, it would be good to study how other rust crypto libraries such as RustCrypto, Graviola, Kryoptic, handle this.

ounsworth:
Tamish Dahiya has volunteered to work on this.

tad-fr:
**RustCrypto**
Maintainer of the Zeroize crate and uses the API extensively across all repos. 
As I understand, the idea is to use a "non-optimised" function to ensure proper zeroing behaviour. It's small enough for a quick read and is implemented [here](https://github.com/RustCrypto/utils/blob/master/zeroize/src/barrier.rs). This function is then wrapped in convenience methods that are exposed via the Zeroize trait.
It does provide some boilerplate for things like arrays and vectors of types that implement Zeroize but nothing too generic; I suppose it would be quite hard to generalize. There is also a derive macro which calls zeroize on all members.
Implementing drops via the Zeroize does provide access to the aforementioned "non-optimised" function which seems like a good approach. The bubbling of the Zeroize as a consequence also seems desirable.

The following also use zeroize copying the RustCrypto pattern:
**Rustls** : https://github.com/rustls/rustls
**dalek-cryptography** : https://github.com/dalek-cryptography
**rust-hpke**: https://github.com/rozbb/rust-hpke

**libcrux**
In the process of moving to zeroize: [pull req](https://github.com/cryspen/libcrux/pull/1332)

**Graviola**
Just implements Drop where appropriate: no marker traits or generic implementations.
Custom infrastructure to zeroize. [Implementation](https://github.com/ctz/graviola/blob/main/graviola/src/low/generic/zeroise.rs).
Implements memset(ptr, 0, len) equivalent in assembly to avoid optimisation and calls it when required. [x86 here](https://github.com/ctz/graviola/blob/main/graviola/src/low/x86_64/cpu.rs#L56).

**ring**
[Does not zeroize](https://github.com/briansmith/ring/blob/7d3dbbb41f19f130ce15a84d69ee28748836b1bb/src/agreement.rs#L297) at all yet.

npajkovsky:
The Zeroize is a good choice to start with. We can harden BC-Rust with supply-chain security
tools like `cargo vet` or `cargo crev` in case something goes sideways.

[...]

PR #36, opened 2026-06-21
Replace Secret's Drop supertrait with ZeroizeOnDrop

tad-fr:
### Description

First pass integrating the zeroize crate. 
Currently only the structs that are already implementing Secret are refactored. The intention is that it allows the PR to be relatively small and also settle on the pattern to be followed before full propagation. 

### Issue

Relates to: #32


crypto/core/src/key_material.rs
```diff
@@ -146,8 +148,6 @@ pub trait KeyMaterialTrait {
 
     fn is_full_entropy(&self) -> bool;
 
-    fn zeroize(&mut self);
-
```
ounsworth: I still think that KeyMaterial should still expose an API to force it to zeroize itself mid-usage, for example if the user want to make sure that it was fully zeroized before writing a smaller key (or a key of unknown size) into it and make sure that none of the original key material is still in the unused part of the buffer.

crypto/core/src/traits.rs
```diff
@@ -387,27 +376,32 @@ impl SecurityStrength {
     }
 }
 
+impl DefaultIsZeroes for SecurityStrength {}
+
+impl Default for SecurityStrength {
+    fn default() -> Self {
+        Self::None
+    }
+}
```
ounsworth: Can you explain this change?
What is the purpose of adding a `::default()` to SecurityStrenght?
(I'm not necessarily saying that it's wrong, I just don't see why this would be used)

crypto/core/src/traits.rs
```diff
@@ -418,10 +412,7 @@ pub trait RNG: Default {
 
 /// A trait that forces an object to implement a zeroizing Drop() as well as Debug and Display that
 /// will not log the sensitive contents, even in error or crash-dump scenarios.
-#[allow(drop_bounds)] // Since rust auto-implements Drop, there's a lint that explicitly bounding on Drop is useless.
-// I disagree because I want to force things that are secrets to manually implement Drop that zeroizes the data.
-// So I'm turning off this lint.
-pub trait Secret: Drop + Debug + Display {}
+pub trait Secret : ZeroizeOnDrop + Debug + Display {}
```
ounsworth: 👍

crypto/core/tests/key_material_tests.rs
```diff
@@ -173,7 +172,12 @@ mod test_key_material {
     #[test]
     fn zeroize() {
         let mut key = KeyMaterial256::from_bytes(&DUMMY_KEY[..32]).unwrap();
```
ounsworth: I think we should put this API back, which means this test will need to be run twice; once for the `.zeroize()` and once for the `drop_in_place()`

crypto/mldsa/src/matrix.rs
```diff
@@ -6,9 +6,10 @@ use crate::mldsa::H;
 use crate::polynomial::Polynomial;
 use bouncycastle_core::traits::XOF;
 use core::ops::{Index, IndexMut};
+use zeroize::ZeroizeOnDrop;
 
 /// A matrix over the ML-DSA ring.
-#[derive(Clone)]
+#[derive(Clone, ZeroizeOnDrop)]
```
ounsworth, line 12: This shouldn't need to be Zeroized because Polynomial is where the secret data is, and it is where we should impl ZeroizeOnDrop.

crypto/mldsa/src/matrix.rs
```diff
@@ -62,7 +63,7 @@ impl<const k: usize, const l: usize> Matrix<k, l> {
 // Technically all matrices and some vectors are only part of the public key and might not need to be zeroized,
 // but I'll leave it zeroizing for now and leave this as a potential future optimization.
 
-#[derive(Clone)]
+#[derive(Clone, ZeroizeOnDrop)]
```
ounsworth, line 66: This shouldn't need to be Zeroized because Polynomial is where the secret data is, and it is where we should impl ZeroizeOnDrop.

crypto/mlkem/src/matrix.rs
```diff
@@ -3,34 +3,34 @@
...
-use crate::mlkem::{N, q};
+use crate::mlkem::{q, N};
 use crate::polynomial;
-use crate::polynomial::Polynomial;
+use crate::polynomial::{Polynomial};
 
-#[derive(Clone)]
+use zeroize::ZeroizeOnDrop;
+
+#[derive(Clone, ZeroizeOnDrop)]
```
ounsworth, line 12: This shouldn't need to be Zeroized because Polynomial is where the secret data is, and it is where we should impl ZeroizeOnDrop.

crypto/mlkem/src/matrix.rs
```diff
@@ -81,10 +81,9 @@ impl<const k: usize, const l: usize> Matrix<k, l> {
 // Technically all matrices and some vectors are only part of the public key and might not need to be zeroized,
 // but I'll leave it zeroizing for now and leave this as a potential future optimization.
 
-#[derive(Clone)]
-pub(crate) struct Vector<const k: usize> {
-    pub(crate) vec: [Polynomial; k],
-}
+
+#[derive(Clone, ZeroizeOnDrop)]
```
ounsworth, line 86: This shouldn't need to be Zeroized because Polynomial is where the secret data is, and it is where we should impl ZeroizeOnDrop.

crypto/mlkem/src/mlkem_keys.rs
```diff
@@ -683,75 +596,72 @@ impl<
     }
 }
 
+impl<const k: usize, PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>, const SK_LEN: usize, const PK_LEN: usize>
+    ZeroizeOnDrop for MLKEMPrivateKey<k, PK, SK_LEN, PK_LEN> {}
```
ounsworth: Could you explain for my understanding why here you provide an empty impl, but in other places you use the #derive macro?

crypto/mlkem/Cargo.toml
```diff
@@ -8,14 +8,14 @@ bouncycastle-core.workspace = true
 bouncycastle-sha3.workspace = true
 bouncycastle-rng.workspace = true
 bouncycastle-utils.workspace = true
+zeroize = { version = "1.9.0", features = ["zeroize_derive"] }
```
ounsworth: Can you please add this to the workspace dependencies in the top-level cargo.toml, so that when we eventually want to bump the version, we only need to do it in one place and not all over?

I'm fairly new to cargo, but I think you can do version = "1.9", and that gives it permission to take 1.9.1, 1.9.2, etc when they become available?

crypto/mlkem-lowmemory/Cargo.toml
```diff
@@ -8,13 +8,13 @@ bouncycastle-core.workspace = true
 bouncycastle-sha3.workspace = true
 bouncycastle-rng.workspace = true
 bouncycastle-utils.workspace = true
+zeroize = { version = "1.9.0", features = ["zeroize_derive"] }
```
ounsworth: Can you please add this to the workspace dependencies in the top-level cargo.toml, so that when we eventually want to bump the version, we only need to do it in one place and not all over?

I'm fairly new to cargo, but I think you can do version = "1.9", and that gives it permission to take 1.9.1, 1.9.2, etc when they become available?

ounsworth (changes requested):

.github/workflows/publish_doc_benches_to_ghpages.yaml
npajkovsky, line 52: If it's not useful, it should be removed in a separate PR.

npajkovsky, line 73: ditto.

crypto/core/src/key_material.rs
npajkovsky, line 545: Why did you suddenly change min/max? If you want to have a prompt review, I suggest sticking with one topic, which is at this point Zeroize. 

It's also a good idea to lay commits in reasonable chunks, for example
```
crypto/core
crypto/mlsda
crypto/mlkem
...
```

crypto/core/src/key_material.rs
```diff
@@ -204,6 +204,14 @@ pub enum KeyType {
...
 }
 
+impl DefaultIsZeroes for KeyType {}
+
+impl Default for KeyType {
+    fn default() -> Self {
+        KeyType::Zeroized
+    }
+}
+
```
npajkovsky:
```patch
-#[derive(Clone, Copy, Debug, Eq, PartialEq)]
+#[derive(Clone, Copy, Debug, Eq, PartialEq, Default)]
 pub enum KeyType {
+    #[default]
     /// The KeyMaterial is zeroized and MUST NOT be used for any cryptographic operation in this state.
     Zeroized,

@@ -206,12 +206,6 @@ pub enum KeyType {

 impl DefaultIsZeroes for KeyType {}

-impl Default for KeyType {
-    fn default() -> Self {
-        KeyType::Zeroized
-    }
-}
```

crypto/core/src/key_material.rs
npajkovsky: This is handy. Please keep it.

crypto/mldsa/Cargo.toml
```diff
@@ -9,14 +9,14 @@ bouncycastle-sha2.workspace = true
 bouncycastle-sha3.workspace = true
 bouncycastle-rng.workspace = true
 bouncycastle-utils.workspace = true
+zeroize = { version = "1.9.0", features = ["zeroize_derive"] }
```
npajkovsky: Should be part of workspace dep.

crypto/mldsa-lowmemory/Cargo.toml
```diff
@@ -9,14 +9,14 @@ bouncycastle-sha2.workspace = true
 bouncycastle-sha3.workspace = true
 bouncycastle-rng.workspace = true
 bouncycastle-utils.workspace = true
+zeroize = { version = "1.9.0", features = ["zeroize_derive"] }
```
npajkovsky: dittto.

crypto/mlkem/src/aux_functions.rs
```diff
@@ -22,9 +23,10 @@ pub(crate) fn expandA<const k: usize>(rho: &[u8; 32]) -> Matrix<k, k> {
 
 /// Algorithm 5 ByteEncode_d(𝐹)
 /// Encodes an array of 𝑑-bit integers into a byte array for 1 ≤ 𝑑 ≤ 12.
-/// Input: integer array 𝐹 ∈ ℤ_M^256 , where 𝑚 = 2^𝑑 if 𝑑 < 12, and 𝑚 = 𝑞 if 𝑑 = 12.
-/// Output: byte array 𝐵 ∈ 𝔹32𝑑 .
-pub(crate) fn byte_encode<const d: usize, const PACK_LEN: usize>(F: &Polynomial) -> [u8; PACK_LEN] {
+/// Input: integer array 𝐹 ∈ ℤ_M^256, where 𝑚 = 2^𝑑 if 𝑑 < 12, and 𝑚 = 𝑞 if 𝑑 = 12.
+/// Output: byte array 𝐵 ∈ 𝔹32𝑑.
+/// Note: this is exposed publicly only for testing purposes and there is no good reason to use it in production code.
+pub fn byte_encode<const d: usize, const PACK_LEN: usize>(F: &Polynomial) -> [u8; PACK_LEN] {
```
npajkovsky, line 28: Is there any driver for making the function public?

crypto/mlkem/src/aux_functions.rs
```diff
@@ -65,7 +66,8 @@ pub(crate) fn byte_encode<const d: usize, const PACK_LEN: usize>(F: &Polynomial)
 /// Decodes a byte array into an array of 𝑑-bit integers for 1 ≤ 𝑑 ≤ 12.
 /// Input: byte array 𝐵 ∈ 𝔹32𝑑 .
 /// Output: integer array 𝐹 ∈ ℤ256 , where 𝑚 = 2𝑑 if 𝑑 < 12 and 𝑚 = 𝑞 if 𝑑 = 12.
-pub(crate) fn byte_decode<const d: usize, const PACK_LEN: usize>(B: &[u8; PACK_LEN]) -> Polynomial {
+/// Note: this is exposed publicly only for testing purposes and there is no good reason to use it in production code.
+pub fn byte_decode<const d: usize, const PACK_LEN: usize>(B: &[u8; PACK_LEN]) -> Polynomial {
```
npajkovsky, line 70: ditto.

crypto/mlkem/src/aux_functions.rs
```diff
@@ -85,7 +87,8 @@ pub(crate) fn byte_decode<const d: usize, const PACK_LEN: usize>(B: &[u8; PACK_L
 /// Takes a 32-byte seed and two indices as input and outputs a pseudorandom element of 𝑇𝑞.
 /// Input: byte array 𝐵 ∈ 𝔹34 . ▷ a 32-byte seed along with two indices
 /// Output: array 𝑎_hat ∈ ℤ256 ▷ the coefficients of the NTT of a polynomial
-pub(crate) fn sample_ntt(rho: &[u8; 32], nonce: &[u8; 2]) -> Polynomial {
+/// Note: this is exposed publicly only for testing purposes and there is no good reason to use it in production code.
+pub fn sample_ntt(rho: &[u8; 32], nonce: &[u8; 2]) -> Polynomial {
```
npajkovsky, line 91: ditto.

crypto/mlkem/src/aux_functions.rs
npajkovsky, line 184: All formatting issues were fixed `upstream/release/0.1.2alpha`. Try rebase please.

crypto/mlkem/src/lib.rs
```diff
@@ -156,15 +156,13 @@
 #[allow(unused_imports)]
 use bouncycastle_core::key_material::KeyMaterialTrait;
 
-// todo -- re-run mutants
-
-// todo -- crucible tests
-
-mod aux_functions;
+// Exposed only for testing purposes
+pub mod aux_functions;
```
npajkovsky, line 158: no need to be pub

crypto/core/src/key_material.rs
tad-fr: This is a result of targeting the main branch rather than the release/0.1.2alpha branch. The diff contained a number of changes unrelated to the zeroize topic at hand. My apologies for wasting your time with this.

crypto/core/src/key_material.rs
tad-fr: 😄

crypto/core/src/traits.rs
tad-fr: Yes, of course. Deriving ZeroizeOnDrop is essentially equivalent to implementing a Drop() where zeroize() is called all fields. For the types and enums that do not have a such an implementation, we can implement DefaultIsZeroes and Default. This basically provides the value that the zeroize() sets. 

The alternative would be to not derive ZeroizeOnDrop and rater implement a Drop() ourselves which sets SecurityStrength and KeyType to the appropriate values.

crypto/core/src/key_material.rs
tad-fr: Will do. I chose to derive the zeroize() as well so we get the benefit of the optimization_barrier() from the Zeroize crate.

crypto/core/tests/key_material_tests.rs
tad-fr: Sounds good.

crypto/mldsa/src/matrix.rs
tad-fr: Deriving ZeroizeOnDrop requires all fields to implement it as well. I chose to derive it on these types with the understanding that rust zero-cost abstractions would result in no additional overhead. I could be wrong of course in which case we could remove the derivations on MLDSAPrivateKey,Matrix and Vector types and implement a Drop() for MLDSAPrivateKey.

crypto/mlkem/src/mlkem_keys.rs
tad-fr: I noticed the existing Drop() implementation had some additional logic in the form of an if statement. I tried to take the prudent approach as a derived Drop() would not have the same check.

crypto/mlkem/src/aux_functions.rs
tad-fr: I have rebased onto the correct branch, changes unrelated to zeroize should not interfere anymore.

crypto/mlkem/src/matrix.rs
tad-fr: Same reason as mldsa/matrix

crypto/mlkem/src/matrix.rs
tad-fr: Same reason as mldsa/matrix

crypto/core/src/key_material.rs
ounsworth: Perfect. I see that `key.zeroize();` is still available, and still shows in the crate docs. Great!

crypto/core/src/traits.rs
ounsworth: So I'm understanding this to be basically a mechanical requirement of implementing ZeroizeOnDrop. Good. Thanks for explaining.

crypto/mldsa/src/matrix.rs
ounsworth: I think I maybe see your point: you've tagged `struct MLDSAPrivateKey` with `#[derive(ZeroizeOnDrop)]`, which I think makes sense for optics, but then in requires that all of its members (Vector, in this case) also impl ZeroizeOnDrop.

But I notice that you have not impl'd ZeroizeOnDrop for `struct MLKEMPRivateKey`, which I assume you can't because that contains an MLKEMPublicKey member, which has no reason to be zeroized.

Which brings up another point:
I'm not yet sure that this is good or bad; I'm just thinking out loud here.
The full struct is:

```rust
pub struct MLDSAPrivateKey<
    const k: usize,
    const l: usize,
    const eta: usize,
    const SK_LEN: usize,
    const PK_LEN: usize,
> {
    rho: [u8; 32],
    K: [u8; 32],
    tr: [u8; 64],
    s1_hat: Vector<l>,
    s2_hat: Vector<k>,
    t0_hat: Vector<k>,
    seed: Option<KeyMaterial<32>>,
}
```

Of those, K, s1, s2, t0, seed are secret data, but rho and tr are the public key and don't strictly need to be zeroized. In this particular case, zeroizing an extra 64 bytes is not a big deal, but it does mean that this pattern where all members also need to impl ZeroizeOnDrop might not give us fine-enough contral if a secret object contains a large non-secret object.

And you actually have an example of that: `crypto/mlkem/src/mlkem_keys.rs` in your PR has:

```rust
#[derive(Clone)]
pub struct MLKEMPrivateKey<
    const k: usize,
    PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>,
    const SK_LEN: usize,
    const PK_LEN: usize,
> {
    s_hat: Vector<k>,
    ek: PK,
    pk_hash: [u8; 32],
    z: [u8; 32],
    seed_d: Option<[u8; 32]>,
}
```

It does not impl ZeroizeOnDrop, probably because MLKEMPublicKeyInternalTrait does not, so this won't compile, but absolutely `z` and `seed_d` are secret and need to be zeroized. Is there another way to do this more surgically? For example, instead of using the derive macro, can we impl ZeroizeOnDrop manually and delete only the things that need to be deleted?

crypto/mlkem/src/mlkem_keys.rs
ounsworth: I don't understand. How does an empty impl help with that?

crypto/mlkem/src/aux_functions.rs
ounsworth: Did you read the comment on it? :wink: 

This was not a change made by Tamish -- it is unrelated to this change set; it just got pulled in when I rebased from `main` to `release/0.1.2alpha`.

Basically, the [crucible harness](https://github.com/symbolicsoft/crucible) needs direct access to this function so that it can do more invasive correctness testing than what you can do via the regular public APIs.
I agree that this really shouldn't be public; but making it public in exchange for better test coverage seems like an ok tradeoff?

crypto/mlkem/src/aux_functions.rs
ounsworth: ditto.

crypto/mlkem/src/aux_functions.rs
ounsworth: ditto.

crypto/mlkem/src/lib.rs
ounsworth: ditto.

ounsworth:
@tad-fr I made some cleanup changes to your branch. I don't know how to push those changes back to your origin (there's probably a way in git, and maybe I'll have time to google it later today), so for now it's here:

https://github.com/ounsworth/bc-rust/tree/tad-fr-secret-zeroize

ounsworth:
@tad-fr , @npajkovsky I have Resolved most of the discussions on this thread -- this is getting close to being ready to merge, but there are a few still open. The big one is how to deal with structs where only part of the content is secret and needs to be zeroized. Can we impl ZeroizeOnDrop manually instead of via a derive macro?

crypto/mldsa/src/matrix.rs
npajkovsky: Is there any reason to have pub key in priv key struct? Those can be separate structs and passed around as tuples.

@ounsworth yes, you're right. The Derive is good only when you know that you want to securely clean, for those who need a partial cleaning, impl ZeroizeOnDrop must be manually implemented.

crypto/mlkem/src/mlkem_keys.rs
tad-fr: Since MLKEMPrivateKey implements Secret, It requires a ZeroizeOnDrop implementation as well; In much the same fashion as we have an empty Secret Impl. 

When we derive ZeroizeOnDrop, this is exactly what the [macro is producing](https://github.com/RustCrypto/utils/blob/2afc28d5bfc4538296d7b21394892e4c46b612a6/zeroize_derive/src/lib.rs#L416).
If we were to derive it on MLKEMPrivateKey, which is defined as: 

```rust
pub struct MLKEMPrivateKey<
    const k: usize,
    PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>,
    const SK_LEN: usize,
    const PK_LEN: usize,
> {
    s_hat: Vector<k>,
    ek: PK,
    pk_hash: [u8; 32],
    z: [u8; 32],
    seed_d: Option<[u8; 32]>,
}
```

the macro would produce:

```rust
impl<
    const k: usize,
    PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>,
    const SK_LEN: usize,
    const PK_LEN: usize,
> ZeroizeOnDrop for MLKEMPrivateKey<k, PK, SK_LEN, PK_LEN> {}

impl<
    const k: usize,
    PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>,
    const SK_LEN: usize,
    const PK_LEN: usize,
> Drop for MLKEMPrivateKey<k, PK, SK_LEN, PK_LEN>
{
    fn drop(&mut self) {
        self.s_hat.zeroize();
        self.ek.zeroize();
        self.pk_hash.zeroize();
        self.z.zeroize();
        self.seed_d.zeroize();
    }
}
```


Since the preexisting Drop() definition did not match the macro output, I chose to not derive it and rather just added the empty impl block. 

Although you could argue I should've updated the fill calls with zeroize() => Which I have just thought of now. 
This would alter the Drop as such: 

```rust
impl<
    const k: usize,
    PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>,
    const SK_LEN: usize,
    const PK_LEN: usize,
> Drop for MLKEMPrivateKey<k, PK, SK_LEN, PK_LEN>
{
    fn drop(&mut self) {
        // s_hat, has its own zeroizing drop
        self.pk_hash.zeroize();
        self.z.zeroize();
        if self.seed_d.is_some() {
               self.seed_d.zeroize();
        }
    }
}
```

-----------------------------------------------------------------------

All of this has made me realize another point: Since our Secret now has ZeroizeOnDrop as a supertrait, ZeroizeOnDrop itself does not have Drop as a supertrait. [ZeroizeOnDrop is just a marker trait.](https://github.com/RustCrypto/utils/blob/2afc28d5bfc4538296d7b21394892e4c46b612a6/zeroize/src/lib.rs#L27) This means the compiler no longer enforces the implementation of a Drop(). The developer must now be aware of the above and remember to implement a Drop with the empty ZeroizeOnDrop impl. This does not seem very developer friendly. 

I see 2 resolutions as of now: 
- Add Drop back to Secret: I cannot see any obvious downside with this?
- Add test coverage for Drop

Curious to hear your opinions.

crypto/mldsa/src/matrix.rs
tad-fr: Yes indeed the necessity is to have an impl ZeroizeOnDrop and an impl Drop. This can be done via a marco or manually. 

I feel the detailed explanation for the mlkem [below](https://github.com/bcgit/bc-rust/pull/36#discussion_r3460514265) covers this as well. Please let me know if i can provide further clarification.

ounsworth: > Is there any reason to have pub key in priv key struct? Those can be separate structs and passed around as tuples.

The simple reason is that FIPS 203 defines the private key as containing the public key and exporting the private key into the NIST encoding requires embedding the public key inside the private key.

The more complex reason is that you cannot use an ML-KEM private key by itself without the public key because Decapsulation (aka decryption) requires both the private key and the public key, which is why FIPS 203 includes the pub inside the priv.

crypto/mlkem/src/mlkem_keys.rs
ounsworth: Hmm. Now I'm a bit confused. I thought ZeroizeOnDrop replaced Drop. Why would you need to implement both?

Also, can we not manually impl ZeroizeOnDrop here as:

```rust
impl<
    const k: usize,
    PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>,
    const SK_LEN: usize,
    const PK_LEN: usize,
> ZeroizeOnDrop for MLKEMPrivateKey<k, PK, SK_LEN, PK_LEN>
{
    fn drop(&mut self) {
        // s_hat, has its own zeroizing drop
        self.z.zeroize();
        if self.seed_d.is_some() {
               self.seed_d.zeroize();
        }
    }
}
```

tad-fr: ### To the first part of your query:

I believe ZeroizeOnDrop is not a subtrait of Drop to allow flexibility.

[Simplified scenario:](https://github.com/iqlusioninc/crates/issues/757)
Person A creates a crate with a type that has some secret data inside, say, `SecretKey { inner: SecretDataType }`. They implement `Zeroize` and `Drop`  for `SecretDataType`, and stop at that - the secret data is now safe (to the extend Rust can guarantee it), the `drop()` of `SecretKey` will call the zeroizing `drop()` of `SecretDataType`, so there's no need to define anything for `SecretKey` itself.

Person B uses A's crate and wants to use `Secret<SecretKey>`. But they cannot, because `SecretKey` does not implement `Zeroize` - so they have to wrap it in a newtype and implement an empty `Zeroize` for it. And they still cannot assert in code that `SecretKey` keeps its secret - they have to rely on the docs. 

So what are the responsibilities of A here? Should they define an (empty) `Zeroize` for `SecretKey`? Should they export a `Secret<Box<SecretKey>>`?

ZeroizeOnDrop aims to solve this. A could implement it for `SecretKey`, and `Secret` can have it as a supertrait. It does not provide any methods, but marks crates as safe, allowing for a scalable API. This is the setup we've gone with so far. Further comparison in next section.
 

---------------------------------------------------------------------------------------------

### Review bouncycastles' Secret's supertraits:

Adding Drop as a supertrait would be counterproductive. Since ZeroizeOnDrop is a marker trait, its the best approach for us in my opinion.

Using hypothetical MLKEMPrivateKey as an example to illustrate. The following might not match the code.
```rust
pub struct MLKEMPrivateKey {
    s_hat: Vector<k>,
}
```
```rust
pub struct Vector {
    vec: [Polynomial; k],
}
```
```rust
pub struct Polynomial {  
    coeffs: [i16; N],
}
```

### Scenario 1 : Secret is a subtrait of Drop only (Setup before this PR)
`Secret: Drop + Debug + Display {}`
To implement Secret on MLKEMPrivateKey we need to trust that the composed types Zeroize properly. This works but the behavior needs to be document instead of the compiler enforcing it. But this allows us to skip Vector boilerplate.
```rust
impl Drop for MLKEMPrivateKey {
    fn drop(&mut self) {
        s_hat.zeroize();
        }
    }
}
```
```rust
impl Drop for Polynomial {
    fn drop(&mut self) {
        coeffs.zeroize();
        }
    }
}
```
 
### Scenario 2 : Secret is a subtrait of Drop AND ZeroizeOnDrop 
`Secret: ZeroizeOnDrop + Drop + Debug + Display {}`
To implement Secret on MLKEMPrivateKey we need to implement both for all types. The compiler is now aware of Zeroize safety but we have empty Drop impl for intermediate type Vector.
```rust
impl ZeroizeOnDrop for MLKEMPrivateKey {}

impl Drop for MLKEMPrivateKey {
    fn drop(&mut self) {
        s_hat.zeroize();
        }
    }
}
```
```rust
impl ZeroizeOnDrop for Vector {}

impl Drop for Vector {}
```

```rust
impl ZeroizeOnDrop for Polynomial {}

impl Drop for Polynomial {
    fn drop(&mut self) {
        coeffs.zeroize();
        }
    }
}
```
 
### Scenario 3 : Secret is a subtrait of ZeroizeOnDrop only (Current Setup)
`Secret: ZeroizeOnDrop + Debug + Display {}`

To implement Secret on MLKEMPrivateKey we need to implement ZeroizeOnDrop for all types. The compiler is now aware of Zeroize safety.  No empty drops are necessary on types that don't need it. 

```rust
impl ZeroizeOnDrop for MLKEMPrivateKey {}

impl Drop for MLKEMPrivateKey {
    fn drop(&mut self) {
        s_hat.zeroize();
        }
    }
}
```
```rust
impl ZeroizeOnDrop for Vector {}
```

```rust
impl ZeroizeOnDrop for Polynomial {}

impl Drop for Polynomial {
    fn drop(&mut self) {
        coeffs.zeroize();
        }
    }
}
```
 



---------------------------------------------------------------------------------------------

### To the second part of your query:

The Drop trait is not implemented on types by default, [its an additional layer _before_ the field destructors are run](https://doc.rust-lang.org/reference/destructors.html#r-destructors.operation). 

Even if that was the case (drop was implemented for all types by default), we could not use the syntax you have specified. You cannot "override" an existing lifecycle hook or an implicit structural behavior simply by changing a local method declaration; behavior must always be mapped through a discrete trait definition by declaring impl Drop {}. The [`<T as core::ops::Drop>::drop`](https://doc.rust-lang.org/core/ops/drop/trait.Drop.html#tymethod.drop) cannot be overriden from a impl ZeroizeOnDrop. [Rust does not inherit.](https://doc.rust-lang.org/book/ch18-01-what-is-oo.html#inheritance-as-a-type-system-and-as-code-sharing) So in order to zeroize a value, a Drop impl must exist (with or without the zeroize crate) as it does stuff that the default "drop glue" does not.

crypto/mlkem/src/aux_functions.rs
npajkovsky: Ah, it's a mess from git merge-back. That's not wise to do.

crypto/mlkem/src/mlkem_keys.rs
ounsworth: I see. I guess I should have just done the research myself from the beginning and read the docs for the Zeroize crate.

https://docs.rs/zeroize/latest/zeroize/trait.ZeroizeOnDrop.html

I didn't realize that ZeroizeOnDrop was itself just an empty marker trait.

So now I don't really understand the value that we're getting from the intrinsics if it's just an empty marker trait and we have to impl Drop and Zeroize ourselves. Maybe I need to find time to actually play with this PR in my own IDE.

npajkovsky: Maybe we're looking at the problem from the wrong angle. Instead of implementing Zeroize on structs that mix secret and non-secret data — which forces a massive chain of Zeroize impls onto types that don't need it — we can do something better.
Let's define a Secret<T> that requires T: Zeroize:

Let's define struct Secret that requires T: Zerosize

```rs
pub struct Secret<T: Zeroize> {
    inner: Box<T>, // heap-pinned, moves don't copy the bytes
}
// Explicit wrapper returned by expose_secret()
// Makes secret access grep-able and visible in code review
pub struct ExposeSecret<'a, T: Zeroize>(pub &'a T);

impl<T: Zeroize> Secret<T> {
    pub fn new(value: T) -> Self {
        Self {
            inner: Box::new(value),
        }
    }

    // Allocate the destination first, hand the keygen a &mut MaybeUninit<T>, and have it write straight into the heap (or, later, the mlock'd arena). The full T never materializes on the stack.
    pub fn new_with(init: impl FnOnce(&mut MaybeUninit<T>)) -> Self {
        let mut uninit = Box::<T>::new_uninit();
        init(&mut *uninit);
        // SAFETY: `init` fully initialized the value.
        Self { inner: unsafe { uninit.assume_init() } }
    }

    // Only way to access the value — name is intentionally loud
    pub fn expose_secret(&self) -> ExposeSecret<'_, T> {
        ExposeSecret(&self.inner)
    }
}

impl<T: Zeroize> Drop for Secret<T> {
    fn drop(&mut self) {
        self.inner.zeroize();
    }
}

// Redact in logs/debug output
impl<T: Zeroize> fmt::Debug for Secret<T> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "Secret([REDACTED])")
    }
}

impl<T: Zeroize> fmt::Display for Secret<T> {
    fn fmt(&self, f: &mut fmt::Formatter<'_>) -> fmt::Result {
        write!(f, "Secret([REDACTED])")
    }
}
```

Then, the Vector we can either have 

```rs
pub(crate) struct Vector<const k: usize> {
    pub(crate) vec: [Secret<Polynomial>; k],
}
```
or vector can also have a more distinguishing type.
```rs
pub(crate) struct SecretVector<const k: usize> {
    pub(crate) vec: [Secret<Polynomial>; k],
}
```

With that said, we can then have MLKEMPrivateKey defined with secrets

```rs
pub struct MLKEMPrivateKey<
    const k: usize,
    PK: MLKEMPublicKeyInternalTrait<k, PK_LEN>,
    const SK_LEN: usize,
    const PK_LEN: usize,
> {
    s_hat: Vector<k>, // or SecretVector
    ek: PK,
    pk_hash: Secret<[u8; 32]>,
    z: Secret<[u8; 32]>,
    seed_d: Option<Secret<[u8; 32]>>,
}
```

That would eliminate Zeroize for MLKEMPrivateKey, and skip pub key entirely.

The story about handling secrets does not end with clearing them. There are two other requirements that have to be done.

1) Swap to disk 
2) Timing side-channels in comparison

It will be much easier to handle those points if the secret is wrapped into own type.

Note, it requires a heap, which we don't have now. 

1) https://benma.github.io/2020/10/16/rust-zeroize-move.html

npajkovsky: Uff, we really need an allocator

```rs
#![no_std]
#![no_main]

use core::fmt::{self, Write};

use zeroize::{Zeroize, ZeroizeOnDrop};

// --- no_std plumbing --------------------------------------------------------

#[link(name = "c")]
unsafe extern "C" {
    fn write(fd: i32, buf: *const u8, count: usize) -> isize;
    fn exit(status: i32) -> !;
}

#[panic_handler]
fn panic(_info: &core::panic::PanicInfo) -> ! {
    unsafe { exit(1) };
}

// libcore's eh_frame references this even with panic = "abort"; it is never called
#[unsafe(no_mangle)]
pub extern "C" fn rust_eh_personality() {}

struct Stdout;

impl fmt::Write for Stdout {
    fn write_str(&mut self, s: &str) -> fmt::Result {
        let buf = s.as_bytes();
        let mut written = 0;
        while written < buf.len() {
            let n = unsafe { write(1, buf[written..].as_ptr(), buf.len() - written) };
            if n < 0 {
                return Err(fmt::Error);
            }
            written += n as usize;
        }
        Ok(())
    }
}

macro_rules! println {
    () => { let _ = <$crate::Stdout as core::fmt::Write>::write_str(&mut $crate::Stdout, "\n"); };
    ($($arg:tt)*) => {{ core::writeln!($crate::Stdout, $($arg)*).ok(); }};
}

#[derive(Debug)]
struct EncryptionKey([u8; 4]);

impl Drop for EncryptionKey {
    fn drop(&mut self) {
        println!("Pointer when zeroing: {:p}", self.0.as_ptr());
        self.0.zeroize();
        println!("Zeroed (before Box::drop): {:?}", self.0);
    }
}

fn get_encryption_key() -> EncryptionKey {
    let key = EncryptionKey(*b"AKey");
    println!("Pointer at creation: {:p}", key.0.as_ptr());
    key
}

#[unsafe(no_mangle)]
pub extern "C" fn main(_argc: i32, _argv: *const *const u8) -> i32 {
    let encryption_key = get_encryption_key();
    let ptr = encryption_key.0.as_ptr();

    println!("Pointer when using: {:p}", encryption_key.0.as_ptr());
    println!("Using key to encrypt stuff: {:?}", encryption_key);

    println!("About to drop.");
    drop(encryption_key);
    println!("Dropped.");

    println!("Memory {:p}: {:?}", ptr, unsafe {
        core::slice::from_raw_parts(ptr, 4)
    });

    0
}
```
```console
Pointer at creation: 0x16f66615c
Pointer when using: 0x16f666214
Using key to encrypt stuff: EncryptionKey([65, 75, 101, 121])
About to drop.
Pointer when zeroing: 0x16f6661a8
Zeroed (before Box::drop): [0, 0, 0, 0]
Dropped.
Memory 0x16f666214: [65, 75, 101, 121]
```

ounsworth:
I took a deep-dive into the Zeroize problem, and came up with a different solution: https://github.com/bcgit/bc-rust/pull/57

@tad-fr I appreciate the effort you put in here, and if you're willing to give that PR a look, I would appreciate that too.

tad-fr:
**Replaced by #57** 

@ounsworth The new solution looks very cool and a big improvement on what we have here. I will therefore close this PR.

PR #40, opened 2026-06-23
Added a closure-based do_hazardous_operations()

### LINKED ISSUE #39: Replace KeyMaterial .allow_hazardous_operations() with a closure-based solution
Basic idea:

```rust
 fn do_hazardous_operations<F>(&mut self, mut f: F) -> Result<(), KeyMaterialError>
 where
       F: FnMut(&mut Self) -> Result<(), KeyMaterialError>,
        Self: Sized,
    {
        self.allow_hazardous_operations();
        let ret = f(self);
        self.drop_hazardous_operations();
        ret
    }
 }
```

And then applications can call it like:

```rust
        key.do_hazardous_operations(|key| {
            rng.next_bytes_out(&mut key.mut_ref_to_bytes().unwrap())
                .map_err(|_| KeyMaterialError::GenericError("RNG failed."))?;

            key.key_len = KEY_LEN;
            key.key_type = KeyType::BytesFullEntropy;
            key.security_strength = rng.security_strength();

            Ok(())
        })?;
```

ounsworth:
This PR deletes the problematic `KeyMaterial.allow_hazardous_operations()` and `KeyMaterial.drop_hazardous_operations()` in favour of a more elegant closure-based solution.

I'm submitting this as a draft with only the core idea in it. If this is in the right direction, then my next steps will be:

* Delete `allow_hazardous_operations()` and `.drop_hazardous_operations()`.
* Refactor all test and doc code that uses it.
* Do a pass over the functions in the KeyMaterial struct and trait to make sure it's clear in the docstrings which ones can be hazardous; and also make them throw consistent error messages that reference needing to be called from within a .do_hazardous_operations() closure.

In the first attempt, this sample code runs successfully (I'm pulling it out because github doesn't seem to highlight doctests nicely).

```rust
use bouncycastle_core::key_material::{KeyType, KeyMaterial256, KeyMaterialTrait};
use bouncycastle_core::traits::SecurityStrength;

// Let's create an all-zero key
let mut key = KeyMaterial256::default();
assert_eq!(key.key_type(), KeyType::Zeroized);
assert_eq!(key.security_strength(), SecurityStrength::None);

// Now we want to tell the library that this all-zero key
// is to be used as a 32-byte [KeyType::Seed] at the 256-bit security strength,
// which the library will not allow you to do outside of the hazerdous operations closure.
key.do_hazardous_operations(|key| {
    key.set_key_len(32)?;
    key.set_key_type(KeyType::Seed)?;
    key.set_security_strength(SecurityStrength::_256bit)?;
    Ok(())
}).unwrap();

assert_eq!(key.key_type(), KeyType::Seed);
assert_eq!(key.security_strength(), SecurityStrength::_256bit);
```

### DIFF
--- a/crypto/core/src/key_material.rs
+++ b/crypto/core/src/key_material.rs
@@ -2,10 +2,6 @@
 //! The main purpose is to hold metadata about the contained key material such as the key type and
 //! entropy content to prevent accidental misuse security bugs, such as deriving cryptographic keys
 //! from uninitialized data.
-//!
-//! This object allows several types of manual-overrides, which typically require setting the [KeyMaterial::allow_hazardous_operations] flag.
-//! For example, the raw bytes data can be extracted, or the key forced to a certain type,
-//! but well-designed use of the bc-rust.test library should not need to ever set the [KeyMaterial::allow_hazardous_operations] flag.
 //! The core idea of this wrapper is to keep track of the usage of the key material, including
 //! the amount of entropy that it is presumed to contain in order to prevent users from accidentally
 //! using it inappropriately in a way that could lead to security weaknesses.
@@ -23,19 +19,36 @@
 //!
 //! However, there is a [KeyMaterial::convert_key_type] for cases where the user has more context knowledge than the library.
 //! Some conversions, such as converting a key of type RawLowEntropy into a SymmetricCipherKey, will fail unless
-//! the user has explicitly allowed them via calling allow_hazardous_operations() prior to the conversion.
+//! run inside of a [do_hazardous_operations] closure, see below.
 //!
-//! Examples of hazardous conversions that require allow_hazardous_operations() to be called first:
-//!
-//! * Converting a KeyMaterial of type RawLowEntropy or RawUnknownEntropy into RawFullEntropy or any other full-entropy key type.
-//! * Converting any algorithm-specific key type into a different algorithm-specific key type, which is considered hazardous since key reuse between different cryptographic algorithms is generally discouraged and can sometimes lead to key leakage.
+//! # Security
 //!
 //! Additional security features:
 //!   * Zeroizes on destruction.
 //!   * Implementing Display and Debug to print metadata but not key material to prevent accidental logging.
 //!
+//! # Hazardous Operations
+//!
+//! This object allows several types of manual-overrides, many of which are considered
+//! "hazardous operations" since by definition they are allowing you to bypass checks meant to detect
+//! conditions that could lead to security vulnerabilities.
+//! Consider, for example, that you are reading a symmetric key from somewhere outside the library,
+//! maybe from disk or from another process, but maybe you handed in the wrong variable and instead
+//! handed in an uninitialized (all-zero) buffer.
+//! Since this is a common bug that has catestrophic security implications, the library will normally
+//! check for all-zero KeyMoterial objects and throw an error.
+//! But there will be cases in which you really do need to use an all-zero key, so you can create
+//! one if you do it in hazardous operations mode.
+//!
+//! Examples of hazardous conversions that are required to be run inside of a do_hazardous_operations() closure:
+//!
+//! * Converting a KeyMaterial of type RawLowEntropy or RawUnknownEntropy into RawFullEntropy or any other full-entropy key type.
+//! * Converting any algorithm-specific key type into a different algorithm-specific key type, which is considered hazardous since key reuse between different cryptographic algorithms is generally discouraged and can sometimes lead to key leakage.
+//!
 //! As with all wrappers of this nature, the intent is to protect the user from making silly mistakes, not to prevent expert users from doing what they need to do.
 //! It as always possible, for example, to extract the bytes from a KeyMaterial object, manipulate them, and then re-wrap them in a new KeyMaterial object.
+//!
+//! See [do_hazardous_operations] for documentation and sample code.
 
 use crate::errors::KeyMaterialError;
 use crate::traits::{RNG, Secret, SecurityStrength};
@@ -53,19 +66,20 @@ pub type KeyMaterial512 = KeyMaterial<64>;
 
 /// A helper class used across the bc-rust.test library to hold bytes-like key material.
 /// See [KeyMaterial] for for details, such as constructors.
-pub trait KeyMaterialTrait {
+#[allow(private_bounds)]
+pub trait KeyMaterialTrait: KeyMaterialInternalTrait {
     /// Loads the provided data into a new KeyMaterial of the specified type.
     /// This is discouraged unless the caller knows the provenance of the data, such as loading it
     /// from a cryptographic private key file.
     ///
-    /// This behaves differently on all-zero input key depending on whether [KeyMaterialTrait::allow_hazardous_operations] is set:
+    /// This behaves differently on all-zero input key depending on whether it is run within a [do_hazardous_operations] closure:
     /// if not set, then it will succeed, setting the key type to [KeyType::Zeroized] and also return a [KeyMaterialError::ActingOnZeroizedKey]
     /// to indicate that you may want to perform error-handling, which could be manually setting the key type
     /// if you intend to allow zero keys, or do some other error-handling, like figure out why your RNG is broken.
     /// Note that even if a [KeyMaterialError::ActingOnZeroizedKey] is returned, the object is still populated and usable.
     /// For example, you could catch it like this:
     /// ```
-    /// use bouncycastle_core::key_material::{KeyMaterial256, KeyType, KeyMaterialTrait};
+    /// use bouncycastle_core::key_material::{KeyMaterial256, KeyType, KeyMaterialTrait, do_hazardous_operations};
     /// use bouncycastle_core::key_material::KeyMaterial;
     /// use bouncycastle_core::errors::KeyMaterialError;
     ///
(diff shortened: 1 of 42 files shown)


crypto/core/src/key_material.rs
```diff
@@ -310,6 +309,47 @@ impl<const KEY_LEN: usize> KeyMaterial<KEY_LEN> {
...
+    /// ```
+    // Dev note: This lives on the concrete type rather than on [KeyMaterialTrait] because a generic method
+    // would make the trait non-dyn-compatible, and the trait is used as `&dyn KeyMaterialTrait`
+    // elsewhere (e.g. [KeyMaterialTrait::concatenate], [KeyMaterialTrait::equals]).
+    pub fn do_hazardous_operations<F>(&mut self, f: F) -> Result<(), KeyMaterialError>
+    where
+        F: FnOnce(&mut Self) -> Result<(), KeyMaterialError>,
+    {
+        self.allow_hazardous_operations = true;
+        f(self)?;
```
npajkovsky: That's wrong, and you will have self.allow_hazardous_operations = true due to bail out via ?.

The correct code is 

```rs
let ret = f(self);
self.allow_hazardous_operations = false // Trun off allow_hazardous_operations no matter what happend
ret
```

ounsworth: Fixed.

ounsworth: Otherwise, this looks like the right direction?

crypto/core/src/key_material.rs
```diff
@@ -310,6 +309,47 @@ impl<const KEY_LEN: usize> KeyMaterial<KEY_LEN> {
...
+    // elsewhere (e.g. [KeyMaterialTrait::concatenate], [KeyMaterialTrait::equals]).
+    pub fn do_hazardous_operations<F>(&mut self, f: F) -> Result<(), KeyMaterialError>
+    where
+        F: FnOnce(&mut Self) -> Result<(), KeyMaterialError>,
+    {
+        self.allow_hazardous_operations = true;
+        let ret = f(self);
+        self.allow_hazardous_operations = false;
+        ret
+    }
```
npajkovsky: Thinking about it more. It seems that all hazard ops come from tests. What would be a good idea is to have KeyMaterial without any hazard ops, and implement hazard ops in tests or even publicly, but separately with feature guard.

ounsworth: HKDF uses hazard ops for example to set an input HMAC key of all zeros (as required by the HKDF spec), or to get a mut_ref to the underlying KeyMaterial byte buffer, which has to work that way in the `_out()` mode where the user hands in an existing KeyMaterial struct to be filled with the output key (there is something similar in the KDF and RNG traits).
https://github.com/bcgit/bc-rust/blob/main/crypto/hkdf/src/lib.rs#L562

And I expect that end users will need to use hazard ops for similar things and override what the library thinks this key buffer contains in order to do whatever you need to do. My idea is to make the user think twice about what they are doing with key bytes, but ultimately allow you to do if it you need to, so this is very much a production feature.

crypto/core/src/key_material.rs
```diff
@@ -630,3 +649,149 @@ impl<const KEY_LEN: usize> Drop for KeyMaterial<KEY_LEN> {
...
+/// // Let's create an all-zero key
+/// let mut key = KeyMaterial256::default();
+/// assert_eq!(key.key_type(), KeyType::Zeroized);
+/// assert_eq!(key.security_strength(), SecurityStrength::None);
+///
+/// // Now we want to tell the library that this all-zero key
+/// // is to be used as a 32-byte [KeyType::Seed] at the 256-bit security strength,
+/// // which the library will not allow you to do outside of the hazerdous operations closure.
+/// do_hazardous_operations(&mut key, |key| {
+///     key.set_key_len(32)?;
```
npajkovsky, line 719: We should stick with either with ? or unwarp(), where unwrap() is easier to handle in runnable/buildable doc.

crypto/core/src/key_material.rs
```diff
@@ -630,3 +649,149 @@ impl<const KEY_LEN: usize> Drop for KeyMaterial<KEY_LEN> {
...
+/// Another common usage of hazardous operations is to get a direct mutable reference to the
+/// underlying KeyMaterial byte buffer; for example if you want to copy in key bytes from somewhere else.
+///
+/// ```rust
+/// use bouncycastle_core::key_material::{KeyType, KeyMaterial512, KeyMaterialTrait, do_hazardous_operations};
+/// use bouncycastle_core::traits::SecurityStrength;
+///
+/// // In this example, we initialize a KeyMateriol512 (64 bytes) with only 32 bytes of input.
+/// let mut key = KeyMaterial512::from_bytes_as_type(
+///                                 &[1u8; 32],
```
npajkovsky, line 738: coding style is off.

crypto/core/src/key_material.rs
```diff
@@ -630,3 +649,149 @@ impl<const KEY_LEN: usize> Drop for KeyMaterial<KEY_LEN> {
...
+    KEY: KeyMaterialTrait,
+    F: FnOnce(&mut KEY) -> Result<(), KeyMaterialError>,
+{
+    let allows = key.allows_hazardous_operations();
+
+    key.allow_hazardous_operations();
+    let ret = f(key);
+
+    // to allow nested closures, if this key instance allowed
+    // before entering, then leave it.
```
npajkovsky, line 775: Is there any case where nested hazard ops are allowed?

crypto/core/src/key_material.rs
```diff
@@ -630,3 +632,149 @@ impl<const KEY_LEN: usize> Drop for KeyMaterial<KEY_LEN> {
...
+// generic over the closure type, which would make the trait non-dyn-compatible; the trait is used
+// as `&dyn KeyMaterialTrait` elsewhere (e.g. [KeyMaterialTrait::concatenate], [KeyMaterialTrait::equals]).
+// The toggle itself lives on the module-private [KeyMaterialInternalTrait], so external crates cannot
+// flip the guard by hand and must go through this scoped wrapper (hence `#[allow(private_bounds)]`).
+#[allow(private_bounds)]
+pub fn do_hazardous_operations<KEY, F>(key: &mut KEY, f: F) -> Result<(), KeyMaterialError>
+where
+    KEY: KeyMaterialTrait + ?Sized,
+    F: FnOnce(&mut KEY) -> Result<(), KeyMaterialError>,
+{
```
npajkovsky, line 768: Let me check what can be done about dyn. I'm just playing around with do_hazardous_operations, and have realised that it always returns (), which isn't handy, and it should return T.

```rs
pub fn do_hazardous_operations<KEY, R, F>(key: &mut KEY, f: F) -> Result<R, KeyMaterialError>
where
    KEY: KeyMaterialTrait + ?Sized,
    F: FnOnce(&mut KEY) -> Result<R, KeyMaterialError>,
{
```

crypto/core/src/key_material.rs
ounsworth: Note that the closure must return Result<(), KeyMaterialError>.
See:
https://github.com/bcgit/bc-rust/pull/40/changes#diff-c1b5c5ca88a37348b130b813cc96710b213c0ad2f15c5931f9938b8b1780c578R764

So that means the pattern is to throw any errors generated inside the closure, and then handle them at outside the do_hazardous_operations. This way, you have the option to do one unwrap, ?, or match for the whole closure instead of needing to handle each thing within the closure that could error.

This sample code does have an .unwrap() at the end of the closure.

crypto/core/src/key_material.rs
ounsworth: Ok. Can you fix it, or suggest what the correct style should be?

crypto/core/src/key_material.rs
ounsworth: I don't have one explicitly in mind, but I if a user wants to do this:

```rust
do_hazardous_operations(&mut seed, |seed| {
    seed.set_security_strength(SecurityStrength::_256bit)?;
    MLDSA87::keygen_from_seed(&seed)?;
   Ok(())
});
```
(syntax may not be exactly correct there).

the keygen_from_seed will itself likely use a do_hazardous_operations closure on the same seed object, so that would be a nested closure. Throwing an error about nested closures and forcing the user to figure out how to move that out of the closure just seems like bad ergonomics ... this should just work.

crypto/core/src/key_material.rs
ounsworth: Hmm. I think I'll go ahead and merge this PR the way it is, and maybe you can prepare a new PR that explores this idea?

I see the potential benefit in being able to return a value from the closure, but also, won't this require the user to set a type for R every time they use the closure? Won't that be annoying to use?

ounsworth:
This is ready for review.

The core change is in this commit:
https://github.com/bcgit/bc-rust/pull/40/commits/04c9955a72db21df2b4d3e17ab3ff98af2ca1fe3

(particularly starting [here](https://github.com/bcgit/bc-rust/pull/40/changes/04c9955a72db21df2b4d3e17ab3ff98af2ca1fe3#diff-c1b5c5ca88a37348b130b813cc96710b213c0ad2f15c5931f9938b8b1780c578R696))

The subsequent commits are refactoring the rest of the library to work again.

crypto/core/src/key_material.rs
npajkovsky: I'll fix it after you merge it.

crypto/core/src/key_material.rs
npajkovsky: Sure, correct coding style according to rustfmt is 

```rs
        let mut key =
            KeyMaterial512::from_bytes_as_type(&[1u8; 32], KeyType::BytesFullEntropy).unwrap();
        assert_eq!(key.key_len(), 32);
```

crypto/sha3/src/sha3.rs
```diff
@@ -81,20 +82,34 @@ impl<PARAMS: SHA3Params> SHA3<PARAMS> {
...
-        )?;
-        output_key.drop_hazardous_operations();
-        output_key.truncate(min(&output_key.key_len(), &PARAMS::OUTPUT_LEN).clone())?;
+        key_material::do_hazardous_operations(&mut *output_key, |output_key| {
+            output_key.set_key_type(key_type)?;
+            output_key.set_security_strength(
+                min(&output_security_strength, &SecurityStrength::from_bits(bytes_written * 8)).clone(),
+            )
+        })
+        .expect(
```
npajkovsky, line 106: This part requires a bit of talk. The library should never abort on error, because that could cause a DoS of the app that is using it. It's better if we return an error here.

bc-rust should go the other way, and that's sweep change focusing on removal of excessive use of `.unwrap()` or `.execpt()`

ounsworth: Good suggestion! I have opened #42 and assigned it to you. Are you willing to start this effort?

Although in this particular case, I disagree that this error should be forwarded to the caller since, as explained in the expect comment, if this ever triggered, it would be because of a programming state error inside this function, not due to bad input from the caller. In other words, this should be infallible, so I will argue that it is correct for the library to swallow this error.

Question: do you think you can construct a test that triggers this expect? That would be the only case in which the err should be returned.

npajkovsky: Fair enough.

npajkovsky (approved):

crypto/core/src/key_material.rs
ounsworth: Thanks! Just to be clear, are you planning PR so that we can discuss it before merging? My intuition says that this may end up making the ergonomics worse, but I could be wrong.

npajkovsky: Yes, I'll submit a PR for discussion.

PR #41, opened 2026-06-24
rename mut_ref_to_bytes to ref_to_bytes_mut for idiomatic Rust

npajkovsky:
Rust API naming (RFC 344) places *_mut* as a suffix on the base name so the mutable accessor mirrors its immutable pair, such as as_ref/as_mut. The old `mut_ref_to_bytes` used a *mut_* prefix, breaking that symmetry with ref_to_bytes.

Rename KeyMaterialTrait::mut_ref_to_bytes to ref_to_bytes_mut so the getter pair reads `ref_to_bytes()` / `ref_to_bytes_mut()`. Updates the trait definition and all call sites across core, hkdf, rng, and sha3, the key_material tests, and the INTRODUCTION.md docs. Behaviour is unchanged.


ounsworth (approved):

ounsworth:
Merged. I'll close this once the sync-bot runs.

PR #43, opened 2026-06-25
dyn RNG

ounsworth:
Things that require randomness can now be handed an instance of `&mut dyn RNG`, including allowing users of the library to impl their own RNG and use it, for example, as the entropy source for an ML-DSA keygen.


ounsworth:
@officialfrancismendoza Does https://github.com/bcgit/bc-rust/pull/43/commits/b95fed96608c57cf3a291b1cc1d39917e6024202 address your [first comment here](https://github.com/bcgit/bc-rust/pull/37#issuecomment-4785718327)?

ounsworth:
A note on this change to traits.rs:

OLD
```rust
    fn add_seed_keymaterial(
        &mut self,
        additional_seed: impl KeyMaterialTrait,
```

NEW
```rust
    fn add_seed_keymaterial(
        &mut self,
        additional_seed: &dyn KeyMaterialTrait,
```

It was nice that this was a move and consumed the `additional_seed` in order to make it difficult for the user to accidentally use it twice, unfortunately, having a method that takes a moved KeyMaterialTrait breaks dyn-compatability of the RNG trait, so this has to be a borrow now. See [this comment](https://github.com/bcgit/bc-rust/pull/37#issuecomment-4793041803) for more detail.

officialfrancismendoza:
> @officialfrancismendoza Does [b95fed9](https://github.com/bcgit/bc-rust/commit/b95fed96608c57cf3a291b1cc1d39917e6024202) address your [first comment here](https://github.com/bcgit/bc-rust/pull/37#issuecomment-4785718327)?

@ounsworth yes, this addresses my first comment functionally. However (nitpick), I don’t see a regression test specifically where encaps_from_rng() rejects an insufficient-strength RNG; the tests added seem focused on keygen. Did we want to add that just in case? If not, everything else looks good.

ounsworth:
@officialfrancismendoza Oh that's not a nitpick, that's a great catch! Thanks for double-checking my work carefully!

officialfrancismendoza:
@ounsworth looks good. Just for the record: I made sure to push upstream to source (David's bare metal server), as changes on this public mirror are overwritten on a regular interval. Since that's now reflected in 0.1.2alpha, approving this feature branch.

officialfrancismendoza (approved):
Approved and squash-merged to 0.1.2alpha on source (to reflect on the public mirror here). LGTM!

PR #44, opened 2026-06-25
Adding traits for symmetric ciphers

ounsworth:
This is prep work for @officialfrancismendoza 's ASCON branch.
These are a first-pass at these traits, and will almost certainly be refined as Francis works through ASCON, and subsequently AES.

Francis, feel free to merge to `release/0.1.2alpha` and rebase your ASCON branch on top.


crypto/core/src/traits.rs
```diff
@@ -17,6 +19,230 @@ pub trait Algorithm {
...
+/// as AEADs or stream ciphers may need to stick extra data either at the beginning or end of the ciphertext.
+/// See the documentation of the underlying implementation for more details.
+pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>:
+    Algorithm + Secret
+{
+    #[cfg(std)]
+    /// A one-shot API to encrypt some plaintext with the given key.
+    /// This function returns the ciphertext as a Vec<u8>, and therefore is only available when compiling with std.
+    /// Returns a tuple containing the initialization data and the ciphertext.
+    /// This is not available if building for no_std.
```
officialfrancismendoza: Made a similar comment in another PR, but std isn't a built-in cfg. nothing in this repo defines it (no Cargo [features], .cargo/config.toml, build.rs, or rustflags), so #[cfg(std)] is always false and encrypt/decrypt/aead_encrypt/aead_decrypt are silently stripped from the trait. Use #[cfg(feature = "std")] and add [features] std = [] (probably default = ["std"]) to crypto/core/Cargo.toml, or drop the Vec APIs.

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+        key.allow_hazardous_operations();
+
+        let security_strengths = [
+            SecurityStrength::None,
+            SecurityStrength::_112bit,
+            SecurityStrength::_128bit,
+            SecurityStrength::_192bit,
+            SecurityStrength::_256bit,
+        ];
+        for ss in security_strengths.iter() {
```
officialfrancismendoza, line 82: set_security_strength calls drop_hazardous_operations() on every success and rejects raising strength without the flag (key_material.rs: lines 406–455). 

Since allow_hazardous_operations() is called only once before the loop, the 2nd iteration (None → _112bit) returns HazardousOperationNotPermitted and this .unwrap() panics. It also rejects strengths above the key length, so KEY_LEN < 32 panics regardless. 

Recommend that you re-enable hazardous ops each iteration and bound tested strengths by KEY_LEN*8. (Same pattern in the BlockCipher and AEAD tests.)

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
```
officialfrancismendoza, line 334: Similar to other PR. This loop varies key strength, you are using mac_key in encrypt_out(&mac_key,...).

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+            C::aead_encrypt_out(&key, aad, msg, &mut ct).unwrap();
+        let (nonce2, _ct_bytes_written, _tag) =
+            C::aead_encrypt_out(&key, aad, msg, &mut ct).unwrap();
+        assert_ne!(nonce1, nonce2);
+
+        // error case: KeyMaterial of wrong type
+        let mac_key =
+            KeyMaterial::<KEY_LEN>::from_bytes_as_type(&DUMMY_SEED_512[..KEY_LEN], KeyType::MACKey)
+                .unwrap();
+        match C::encrypt_out(&mac_key, msg, &mut ct) {
```
officialfrancismendoza: If we want to test the wrong key check, wouldn't we want to use aead_encrypt_out?

crypto/core-test-framework/src/symmetric_ciphers.rs
```diff
@@ -0,0 +1,361 @@
...
+        ct[17] ^= 0xFF;
+        match C::aead_decrypt_out(&key, &nonce, aad, &ct[..ct_bytes_written], &tag, &mut pt) {
+            Ok(bytes_written) => {
+                // so it decrypted something, but it had better not match the original plaintext
+                assert_eq!(bytes_written, pt_bytes_written);
+                assert_ne!(&pt[..bytes_written], msg);
+            }
+            Err(SymmetricCipherError::DecryptionFailed) => { /* also ok */ }
+            _ => panic!("Unexpected error"),
+        };
```
officialfrancismendoza, line 273: For Authenticated Encryption with Associated Data (AEAD), modifying ciphertext with unchanged tag must also fail a tag check. Accepting Ok(_) lets a cipher that never authenticate changed ciphertext pass.

Recommend Err(AEADTagCheckFailed) here, to match AAD- and tag-tamper cases below.

crypto/core/src/traits.rs
```diff
@@ -17,6 +19,230 @@ pub trait Algorithm {
...
+    /// A one-shot API to decrypt some ciphertext with the given key.
+    /// This function returns the ciphertext as a Vec<u8>, and therefore is only available when compiling with std.
+    fn aead_decrypt(
+        key: &KeyMaterial<KEY_LEN>,
+        nonce: &[u8; NONCE_LEN],
+        aad: &[u8],
+        ciphertext: &[u8],
+        tag: &[u8; TAG_LEN],
+    ) -> Result<Vec<u8>, SymmetricCipherError>;
+    /// A one-shot API to decrypt some ciphertext with the given key.
```
officialfrancismendoza, line 174: tag: &[u8] drops compile-time length guarantees that aead_encrypt_out and aead_decrypt uses [u8, TAG_LEN] and violates the "prefer &[u8; N]" rule. Also allows tag-truncation forgeries if an implementer compares tag.len() bytes.

Recommend we change to tag: &[u8; TAG_LEN]

crypto/core/src/traits.rs
```diff
@@ -17,6 +19,230 @@ pub trait Algorithm {
...
+    /// This function takes a reference to the output buffer for the plaintext, and is therefore available in no_std.
+    /// See the documentation for the underlying implementation for details on providing a plaintext buffer of sufficient size;
+    /// typically the ciphertext is the same length as the plaintext, but some ciphers may have an expansion factor or require
+    /// extra space for a nonce or tag.
+    /// Returns a tuple containing the initialization data and the number of bytes written to the plaintext buffer.
+    fn aead_decrypt_out(
```
officialfrancismendoza, line 180: tag: &[u8] drops compile-time length guarantees that aead_encrypt_out and aead_decrypt uses [u8, TAG_LEN] and violates the "prefer &[u8; N]" rule. Also allows tag-truncation forgeries if an implementer compares tag.len() bytes.

Recommend we change to tag: &[u8; TAG_LEN]

crypto/core/src/traits.rs
```diff
@@ -17,6 +19,230 @@ pub trait Algorithm {
...
+
+/// The basic functions of a block cipher.
+/// This trait allows for a block cipher to generate initialization data, such as an Initialization Vector (IV) or Counter (CTR)
+/// which is not technically part of the ciphertext, but must be transmitted along with the ciphertext in order for the
+/// recipient to perform successful decryption. The length of the initialization data is specified by the implementing struct
+/// via the `INIT_DATA_SIZE` constant.
+/// In order for these one-shot APIs to be usable securely in all contexts, the init data will be generated
+/// securely by the block cipher implementation and returned along with the ciphertext, and there is no API for the
+/// user to provide the init data. If you require this functionality, see the documentation for the underlying implementation.
+pub trait BlockCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize, const BLOCK_LEN: usize>:
```
officialfrancismendoza, line 82: **Design Question**: Encryption has do_encrypt_final/_out but decryption has no do_decrypt_final. 

Padding modes need a decrypt-side finalization to strip/validate padding. Was this intentional, or should we add it?

crypto/core/src/traits.rs
```diff
@@ -17,6 +19,230 @@ pub trait Algorithm {
...
+    ) -> Result<[u8; BLOCK_LEN], SymmetricCipherError>;
+    /// Decrypts a single block of ciphertext and writes the plaintext to the provided buffer.
+    fn do_decrypt_block_out(
+        &mut self,
+        ciphertext: &[u8; BLOCK_LEN],
+        plaintext: &mut [u8; BLOCK_LEN],
+    ) -> Result<usize, SymmetricCipherError>;
+}
+
+/// The basic functions of an Authenticated Encryption with Addititional Data cipher.
```
officialfrancismendoza, line 130: **Design question**: This forces every AEAD to expose the unauthenticated one-shot encrypt_out/decrypt_out (tag dropped). Problematic for a "fool-proof" API. Should AEAD inherit these, and what are their semantics on an AEAD type?

crypto/core/src/traits.rs
ounsworth: Correct, for now, it's always false.

I've created ticket #49 because I think that choosing a pattern for this is really its own task.

Should we merge this PR the way it is, or fully comment-out that block, or wait until #49 is resolved? I could go any way here.

crypto/core-test-framework/src/symmetric_ciphers.rs
ounsworth: Well spotted.
Actually, I've already solved this problem on `release/0.1.2alpha` via #40 .
Once this PR gets rebased on top of that (or merged into it), this problem will go away.

crypto/core/src/traits.rs
ounsworth: I thought I had added one. Maybe I just thought about adding one. Yes. We will certainly need one.

crypto/core/src/traits.rs
ounsworth: You're making an assumption that that'll be unauthenticated :wink: 
Couldn't an AEAD implement the one-shot encrypt_out by sticking the tag at the end of the ciphertext, and have the corresponding decrypt_out expect that? Like, just silently do the correct AEAD tag stuff without even telling the user that you're doing it.

crypto/core/src/traits.rs
officialfrancismendoza: @ounsworth I think to unblock, we'll merge this PR the way it is (but add a TODO), and then retroactively patch with #49

crypto/core-test-framework/src/symmetric_ciphers.rs
officialfrancismendoza: @ounsworth noted: will pull #40 so I can use their version of hazardous operations

- [x] Use #40 hazardous operations

crypto/core-test-framework/src/symmetric_ciphers.rs
officialfrancismendoza: Note: test both encrypt_out and aead_encrypt_out

crypto/core/src/traits.rs
officialfrancismendoza: - [x] TODO: @officialfrancismendoza add do_decrypt_final

Issue #42, opened 2026-06-25
Sweeping removal of .unwrap() and .expect() in library code


ounsworth:
Nikola requests that we do a sweeping change focusing on removal of excessive use of .unwrap() or .execpt() in order to reduce the surface for the library to abort on error.

So the task for this ticket is to systematically go through all the .unwrap() and .expect() sites within the library and see if any can be removed -- not forwarded to the caller, which is just as bad, but removed by restructuring the code so that the error is structurally eliminated.

ounsworth:
I personally feel that I have been very careful in applying the following philosophy -- in ranked order:

1. The library should return errors to the caller _only_ in instances of bad input (either bad input data or bad state).
2. The library should never forward errors to the caller due to bad programming within the library (for example, internally using a hash in an invalid state, or triggering a hazardous KeyMaterial error outside of a do_haz_ops block). In these cases, the correct behaviour is to use a .unwrap() or .expect() inside the library, document via a comment why this is expected to be infallible, and provide good unit testing to convince ourselves that this is un-triggerable.

So if I have done my job well, I hope there are already very few call sites of .unwrap() or .expect() that can be removed, but I bet there are some.

Issue #23, opened 2026-06-17
API Design: Factory Pattern Friction With Rust


officialfrancismendoza:
## Problem
Rust is a multi-paradigm systems programming language that is not purely object-oriented. In comparison to other Bouncy Castle cryptography libraries that are expressed in native object-oriented languages (C# and Java, for example), Rust is unique given its strict nature.

As part of API design, factories are a contentious issue. In C# and Java counterparts, they are a utility class that creates cryptographic objects from encoded data without the caller needing to know the exact implementation type, which follows a factory pattern. They both use runtime data-driven dispatch (read and parse metadata, find discriminator field, choose constructor based on that value) to determine what specific cryptographic constructor to use.

Rust does **not** naturally model this pattern in the same way as Java or C#. As part of API design, we need to determine how factory-style decoding should be represented:

- Enum-based return types
- Trait objects
- Algorithm-specific constructors only
- Some hybrid approach

The goal is to provide a Rust-idiomatic API while preserving support for decoding algorithm-agnostic encoded data.

mcr:
My experience is with Ruby (duck typing), C, and Rust.  I do tolerate Java, and have never done C#. 
I have seen the Factory pattern all over Java, and I've seldom understood what it was really doing for me.
In Rust, we have a pattern of having "Builders", and I don't find that pattern particularly exciting.  
What I see in many places is that either the author of the application knows EXACTLY what they want, and they spend 50 lines of code convincing the Factory or Builder about this.  Or, they either just don't care (give me the best default).  
Added to this is various levels of nonsense related to how some operator/user of the application wants to configure things.
(The OpenSSL provider interface is very half-baked, has surprising limits/complexities)

The place where I think this is MOST visible is around private keys/signing.  Because the source of the private key is increasingly not a file in a magic directory, but some HSM-like thing (TEE, TPM, USB/Yubico, TCP-connection-to-HSM-appliance).  
The pattern that I think is best is dependency injection via "dyn trait".  
This pattern is pretty much mandatory in Rust to enable sensible testing, and one sees this in the way that Rng are passed in.
OTH, I really hate the limitations that dyn trait creates, causing one to Box<> things one might not have, or resulting in untypable (like, with a keyboard...) type names.  
I hate that I wind up with a dyn trait JUST so that I can write test cases, but any real application is going to pick *one* thing.  Maybe LTO can make me feel less wasteful, but ...

[...]

PR #46, opened 2026-06-26
Removal of .unwrap() and .expect() in library code

npajkovsky:
Most of the .unwrap(), .expect(), panic or unreachable can be tolerated, and this PR fixes
a couple of unsoundness in the library.

Closes https://github.com/bcgit/bc-rust/issues/42

### DIFF
--- a/crypto/factory/src/hash_factory.rs
+++ b/crypto/factory/src/hash_factory.rs
@@ -35,11 +35,6 @@ use bouncycastle_sha2::{SHA224_NAME, SHA256_NAME, SHA384_NAME, SHA512_NAME};
 use bouncycastle_sha3 as sha3;
 use bouncycastle_sha3::{SHA3_224_NAME, SHA3_256_NAME, SHA3_384_NAME, SHA3_512_NAME};
 
-/*** Defaults ***/
-pub const DEFAULT_HASH_NAME: &str = SHA3_256_NAME;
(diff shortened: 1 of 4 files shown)


crypto/factory/src/hash_factory.rs
```diff
@@ -35,11 +35,6 @@ use bouncycastle_sha2::{SHA224_NAME, SHA256_NAME, SHA384_NAME, SHA512_NAME};
 use bouncycastle_sha3 as sha3;
 use bouncycastle_sha3::{SHA3_224_NAME, SHA3_256_NAME, SHA3_384_NAME, SHA3_512_NAME};
 
-/*** Defaults ***/
-pub const DEFAULT_HASH_NAME: &str = SHA3_256_NAME;
-pub const DEFAULT_128BIT_HASH_NAME: &str = SHA3_256_NAME;
-pub const DEFAULT_256BIT_HASH_NAME: &str = SHA3_512_NAME;
-
```
ounsworth, line 42: Hmm. This change will make it hard in the future if we want to make the library defaults driven from a config file or env var, but I guess for now this saves us 3 unwraps per factory. Ok.

(also, I think the whole factory crate needs a big re-visit, see #23)

ounsworth (approved):

ounsworth:
Merged manually.

PR #48, opened 2026-07-01
New feature: Suspendable (serializable state)

### LINKED ISSUE #24: Add a trait to serialize the state of stateful structs
Algorithms that implement the `do_update()`, `do_final()` pattern (hash functions, signatures, ciphers, etc) typically do so because the input could be large and could be received over the network in multiple asynchronous chunks. In some applications, it can be advantageous to be able to serialize the algorithm's internal state to a cache so that it can be resumed (potentially on a different host) when the next chunk of input arrives

This task is to add a trait for serializing an algorithm state to a byte array.

Potentially something similar to:

``` rust
/// Cryptographic algorithms that implement this trait are providing a way to serialize their internal state so that they can be
/// paused and resumed at a later time, typically as part of a `do_update()`, `do_final()` flow.
trait SerializableState<const STATE_LEN: usize> {
     /// Serialize the algorithm state to a byte array; typically used to pause an algorithm in the middle of
     /// a `do_update()`, `do_final()` flow.
     ///
     /// Note that this is not intended as a mechanism for cloning stateful operations.
     /// As such, this function moves `self` so that the struct is no longer usable after its state has been serialized
     /// as an attempt to help prevent bugs and vulnerabilities related to accidental state reuse.
     /// If you are intending to clone the operational state, `.clone()` should be available for this purpose.
     ///
     /// Notes to implementers: 
     ///   * The serialized state MUST include the version of the library that produced it.
    ///    * The serialized state MUST NOT include any private key material
    ///
    /// Security Consideration:
    ///    While the serialized state does not contain the private key material, it must still be treated as sensitive
    ///    since its leakage can still lead to any number of security issues.
   fn serialize_state(self) -> [u8; STATE_LEN];


    /// Create a new instance of an algorithm struct from a serialized state; typically used to pause an algorithm in the middle of
    /// a `do_update()`, `do_final()` flow.
    ///
    /// Note that this is not intended as a mechanism for cloning stateful operations.
    /// As such, this function moves the serialized state array in order to help prevent bugs and vulnerabilities related
    /// to accidental state reuse.
    ///
    /// Notes to implementers: 
    ///    * If ever the serialization format for a given algorithm needs to be changed, this function MUST check the embedded
    ///       library version and throw [SerializationError::IncompatibleVersion] if the serialized state was produced by a version
    ///       of the library before the breaking change. The oldest version accepted MUST be documented in the corresponding docstring.
    ///       
   fn from_serialized_state(state: [u8; STATE_LEN]) -> Result<Self, SerializedStateError>;
}

enum SerializedStateError {
    CorruptState,
    IncompatibleVersion,
}
```

Some extra thoughts to whomever picks up this ticket:
  * As part of addressing this ticket, think through whether the proposed trait above makes sense, add it to bouncycastle-core, and impl it for all stateful algorithms.
  * The requirement to not serialize the private key might force us to redesign some of the algorithm APIs, or maybe add a `from_serialized_state_with_key(state: [u8; STATE_LEN], key: &[u8; KEY_LEN])` or something.

ounsworth:
Closes #24 

This was a feature requested by Simo to be able to pause an operation -- particularly in the middle of a `do_update()` ... `do_update()` ... `do_final()` flow -- stick the serialized crypto operation into a cache, and then resume it later. This is an important feature for implementing TLS loadbalancers that want to be able to cache out the HKDF state and resume it from a different box when the next flight of TCP messages come in.

The only "creative" thing I've added here is that all serialized states are prefixed with a 3-byte semantic version of the library and it's checked on load back in. The idea is that if we ever need to change the serialization of some object, then we can add a guard to refuse to load a serialized object from before the guard (or keep both versions of the deserializer and switch based on the version tag, or whatever ... #futureProofing).

### DIFF
--- a/crypto/core/src/errors.rs
+++ b/crypto/core/src/errors.rs
@@ -1,10 +1,4 @@
 #[derive(Debug)]
-pub enum CoreError {
-    IncompatibleVersion,
-    InvalidData,
-}
-
-#[derive(Debug)]
 pub enum HashError {
     GenericError(&'static str),
     InvalidLength(&'static str),
@@ -76,6 +70,17 @@ pub enum RNGError {
     SecurityStrengthInsufficientForAlgorithm,
 
     KeyMaterialError(KeyMaterialError),
+}
+
+#[derive(Debug)]
+pub enum SerializedStateError {
+    /// The serialized state was produced by a library version incompatible with this one.
+    IncompatibleVersion,
+    /// The serialized state is malformed or corrupt.
+    InvalidData,
+    /// The key supplied to [crate::traits::SerializableKeyedState::from_serialized_state] does not
+    /// match the key the state was created with (it is bound to a different public-key hash `tr`).
+    IncorrectKey,
 }
 
 #[derive(Debug)]
--- a/crypto/core/src/errors.rs
+++ b/crypto/core/src/errors.rs
@@ -78,7 +78,7 @@ pub enum SerializedStateError {
     IncompatibleVersion,
     /// The serialized state is malformed or corrupt.
     InvalidData,
-    /// The key supplied to [crate::traits::SerializableKeyedState::from_serialized_state] does not
+    /// The key supplied to [crate::traits::SuspendableKeyed::from_suspended] does not
     /// match the key the state was created with (it is bound to a different public-key hash `tr`).
     IncorrectKey,
 }
--- a/crypto/core/src/errors.rs
+++ b/crypto/core/src/errors.rs
@@ -78,9 +78,6 @@ pub enum SuspendableError {
     IncompatibleVersion,
     /// The serialized state is malformed or corrupt.
     InvalidData,
-    /// The key supplied to [crate::traits::SuspendableKeyed::from_suspended] does not
-    /// match the key the state was created with (it is bound to a different public-key hash `tr`).
-    IncorrectKey,
 }
 
 #[derive(Debug)]
--- a/crypto/core/src/errors.rs
+++ b/crypto/core/src/errors.rs
@@ -1,4 +1,9 @@
 #[derive(Debug)]
+pub enum CoreError {
+    IncompatibleVersion,
+}
+
+#[derive(Debug)]
 pub enum HashError {
     GenericError(&'static str),
     InvalidLength(&'static str),
--- a/crypto/core/src/errors.rs
+++ b/crypto/core/src/errors.rs
@@ -73,7 +73,7 @@ pub enum RNGError {
 }
 
 #[derive(Debug)]
-pub enum SerializedStateError {
+pub enum SuspendableError {
     /// The serialized state was produced by a library version incompatible with this one.
     IncompatibleVersion,
     /// The serialized state is malformed or corrupt.
--- a/crypto/core/src/errors.rs
+++ b/crypto/core/src/errors.rs
@@ -1,6 +1,7 @@
 #[derive(Debug)]
 pub enum CoreError {
     IncompatibleVersion,
+    InvalidData,
 }
 
 #[derive(Debug)]
--- a/crypto/core/src/serializable_state.rs
+++ b/crypto/core/src/serializable_state.rs
@@ -45,17 +45,12 @@ fn test_cmp_lib_ver() {
     assert_eq!(cmp_lib_ver(&[1, 1, 1], &[1, 1, 1]), 0);
 }
 
-/// A helper for serializing an object's state
+/// Puts the library version into the first three bytes of the state array.
 ///
-/// The state array must have length SERIALIZED_LEN - 3 to account for adding the 3-byte symver tag.
-pub fn add_lib_ver<const SERIALIZED_LEN: usize>(state: &[u8]) -> [u8; SERIALIZED_LEN] {
-    assert_eq!(state.len(), SERIALIZED_LEN - 3);
-
-    let mut out = [0u8; SERIALIZED_LEN];
-    out[..3].copy_from_slice(&LIB_VERSION);
-    out[3..].copy_from_slice(state);
-
-    out
+/// Hands back a slice to the same array, starting after the version tag.
+pub fn add_lib_ver<const SERIALIZED_LEN: usize>(state: &mut [u8; SERIALIZED_LEN]) -> &mut [u8] {
+    state[..3].copy_from_slice(&LIB_VERSION);
+    &mut state[3..]
 }
 
 /// A helper for deserializing an object's state
@@ -67,14 +62,13 @@ pub fn add_lib_ver<const SERIALIZED_LEN: usize>(state: &[u8]) -> [u8; SERIALIZED
 ///
 /// Note that for testability, this will always reject if the serialized state contains a version tag
 /// of `[0,0,0]`.
-pub fn remove_lib_ver<const SERIALIZED_LEN: usize>(
-    state_in: &[u8; SERIALIZED_LEN],
-    state_out: &mut [u8],
+///
+/// Hands back a slice to the same array, starting after the version tag.
+pub fn check_lib_ver<const SERIALIZED_LEN: usize>(
+    state: &[u8; SERIALIZED_LEN],
     not_before: Option<[u8; 3]>,
-) -> Result<usize, CoreError> {
-    assert!(state_out.len() >= SERIALIZED_LEN - 3);
-
-    let ver: [u8; 3] = state_in[..3].try_into().unwrap();
+) -> Result<&[u8], CoreError> {
+    let ver: [u8; 3] = state[..3].try_into().unwrap();
 
     let not_before = not_before.unwrap_or([0, 0, 0]);
 
@@ -85,6 +79,5 @@ pub fn remove_lib_ver<const SERIALIZED_LEN: usize>(
         return Err(CoreError::IncompatibleVersion);
     };
 
-    state_out.copy_from_slice(&state_in[3..]);
-    Ok(SERIALIZED_LEN - 3)
+    Ok(&state[3..])
 }
--- a/crypto/core/src/serializable_state.rs
+++ b/crypto/core/src/serializable_state.rs
@@ -1,0 +1,90 @@
+//! Helper functions for standardizing serialization and deserialization of stateful objects.
+
+use crate::errors::CoreError;
+
+/// The current library version.
+// There is almost certainly a more elegant way to do this.
+pub const LIB_VERSION: [u8; 3] = [0, 1, 2];
+
+/// Compare two library semantic versions in the standard C format:
+/// * if a < b => -1
+///* if a == b => 0
+///* if a > b => 1
+pub fn cmp_lib_ver(a: &[u8; 3], b: &[u8; 3]) -> i8 {
+    if a[0] < b[0] {
+        -1
+    } else if a[0] > b[0] {
+        1
+    }
+    // first component is equal
+    else if a[1] < b[1] {
+        -1
+    } else if a[1] > b[1] {
+        1
+    }
+    // first two components are equal
+    else if a[2] < b[2] {
+        -1
+    } else if a[2] > b[2] {
+        1
+    }
+    // all three components are equal
+    else {
+        0
+    }
+}
+
+#[test]
+fn test_cmp_lib_ver() {
+    assert_eq!(cmp_lib_ver(&[0, 2, 1], &[1, 1, 1]), -1);
+    assert_eq!(cmp_lib_ver(&[2, 1, 1], &[1, 1, 1]), 1);
+    assert_eq!(cmp_lib_ver(&[1, 0, 2], &[1, 1, 1]), -1);
+    assert_eq!(cmp_lib_ver(&[1, 2, 0], &[1, 1, 1]), 1);
+    assert_eq!(cmp_lib_ver(&[1, 1, 0], &[1, 1, 1]), -1);
+    assert_eq!(cmp_lib_ver(&[1, 1, 2], &[1, 1, 1]), 1);
+    assert_eq!(cmp_lib_ver(&[1, 1, 1], &[1, 1, 1]), 0);
+}
+
+/// A helper for serializing an object's state
+///
+/// The state array must have length SERIALIZED_LEN - 3 to account for adding the 3-byte symver tag.
+pub fn add_lib_ver<const SERIALIZED_LEN: usize>(state: &[u8]) -> [u8; SERIALIZED_LEN] {
+    assert_eq!(state.len(), SERIALIZED_LEN - 3);
+
+    let mut out = [0u8; SERIALIZED_LEN];
+    out[..3].copy_from_slice(&LIB_VERSION);
+    out[3..].copy_from_slice(state);
+
+    out
+}
+
+/// A helper for deserializing an object's state
+///
+/// The state_out array must have length at least SERIALIZED_LEN - 3.
+///
+/// Returns the number of bytes written to state_out, or a [CoreError::IncompatibleVersion] if the
+/// serialized state contains a version header earlier than the specified `not_before` version.
+///
+/// Note that for testability, this will always reject if the serialized state contains a version tag
+/// of `[0,0,0]`.
+pub fn remove_lib_ver<const SERIALIZED_LEN: usize>(
+    state_in: &[u8; SERIALIZED_LEN],
+    state_out: &mut [u8],
+    not_before: Option<[u8; 3]>,
+) -> Result<usize, CoreError> {
+    assert!(state_out.len() >= SERIALIZED_LEN - 3);
+
+    let ver: [u8; 3] = state_in[..3].try_into().unwrap();
+
+    let not_before = not_before.unwrap_or([0, 0, 0]);
+
+    if cmp_lib_ver(&ver, &not_before) < 0 {
+        return Err(CoreError::IncompatibleVersion);
+    };
+    if ver == [0, 0, 0] {
+        return Err(CoreError::IncompatibleVersion);
+    };
+
+    state_out.copy_from_slice(&state_in[3..]);
+    Ok(SERIALIZED_LEN - 3)
+}
--- a/crypto/core/src/serializable_state.rs
+++ b/crypto/core/src/serializable_state.rs
@@ -2,54 +2,56 @@
 
 use crate::errors::CoreError;
 
-/// The current library version.
-// There is almost certainly a more elegant way to do this.
-pub const LIB_VERSION: [u8; 3] = [0, 1, 2];
+/// A semantic library version, ordered by `major`, then `minor`, then `patch`.
+///
+/// The field declaration order matters: the derived [`Ord`]/[`PartialOrd`] compare fields
+/// lexicographically in declaration order, which is exactly semantic-version precedence.
+#[derive(Debug, Clone, Copy, PartialEq, Eq, PartialOrd, Ord)]
+pub struct SemVer {
+    pub major: u8,
+    pub minor: u8,
+    pub patch: u8,
+    // A semantic version can often also take a suffix, e.g. "alpha", "beta", "rc1", etc.
+    // We're not going to model that here because it's not useful for versioning serialized states.
+}
 
-/// Compare two library semantic versions in the standard C format:
-/// * if a < b => -1
-///* if a == b => 0
-///* if a > b => 1
-pub fn cmp_lib_ver(a: &[u8; 3], b: &[u8; 3]) -> i8 {
-    if a[0] < b[0] {
-        -1
-    } else if a[0] > b[0] {
-        1
-    }
-    // first component is equal
-    else if a[1] < b[1] {
-        -1
-    } else if a[1] > b[1] {
-        1
-    }
-    // first two components are equal
-    else if a[2] < b[2] {
-        -1
-    } else if a[2] > b[2] {
-        1
-    }
-    // all three components are equal
-    else {
-        0
+impl From<[u8; 3]> for SemVer {
+    fn from(v: [u8; 3]) -> Self {
+        SemVer { major: v[0], minor: v[1], patch: v[2] }
     }
 }
 
+impl From<SemVer> for [u8; 3] {
+    fn from(v: SemVer) -> Self {
+        [v.major, v.minor, v.patch]
+    }
+}
+
+/// The current library version.
+// There is almost certainly a more elegant way to do this.
+pub const LIB_VERSION: SemVer = SemVer { major: 0, minor: 1, patch: 2 };
+
 #[test]
 fn test_cmp_lib_ver() {
-    assert_eq!(cmp_lib_ver(&[0, 2, 1], &[1, 1, 1]), -1);
-    assert_eq!(cmp_lib_ver(&[2, 1, 1], &[1, 1, 1]), 1);
-    assert_eq!(cmp_lib_ver(&[1, 0, 2], &[1, 1, 1]), -1);
-    assert_eq!(cmp_lib_ver(&[1, 2, 0], &[1, 1, 1]), 1);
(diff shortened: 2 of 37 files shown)


ounsworth:
For now, I have added the trait and impl'd it for SHA2.

I'll pause here and give a chance for people to give feedback before I continue with doing the same for:

- [x] SHA3
- [x] HMAC
- [x] HKDF
- [x]  ML-DSA
- [x] ML-DSA_lowmemory

crypto/core/src/serializable_state.rs
```diff
@@ -0,0 +1,83 @@
...
+/// The current library version.
+// There is almost certainly a more elegant way to do this.
+pub const LIB_VERSION: [u8; 3] = [0, 1, 2];
+
+/// Compare two library semantic versions in the standard C format:
+/// * if a < b => -1
+///* if a == b => 0
+///* if a > b => 1
+pub fn cmp_lib_ver(a: &[u8; 3], b: &[u8; 3]) -> i8 {
+    if a[0] < b[0] {
```
npajkovsky:
```rs
let a_ver = a[2] << 16 | a[1] << 8 | a[0]
let b_ver = b[2] << 16 | b[1] << 8 | b[0]

if a_ver < b_ver
   return -1;
else if ...
```

Also, give it a type such as SemVer([u8; 3]), and then `impl Ord` would be a much rusty way. Or even better, 

```rs
struct SemVer(u32);
impl SemVer
   fn new(major: u8, minor: u8, patch: u8) -> Self {
     Self(major << 16 | ...)
   }
```

crypto/sha2/src/sha256.rs
```diff
@@ -306,3 +303,67 @@ impl<PARAMS: SHA2Params> Hash for SHA256Internal<PARAMS> {
...
 }
+
+impl<PARAMS: SHA2Params> SerializableState<108> for SHA256Internal<PARAMS> {
+    fn serialize_state(&self) -> [u8; 108] {
+        let mut out_to_return = [0u8; 108];
+
+        // insert the version tag
+        let out: &mut [u8; 105] = add_lib_ver(&mut out_to_return).try_into().unwrap();
+
+        // state.h: [u32; 8]
```
npajkovsky, line 325: What is state.h?

crypto/sha2/src/sha256.rs
```diff
@@ -306,3 +303,67 @@ impl<PARAMS: SHA2Params> Hash for SHA256Internal<PARAMS> {
...
+        }
+
+        // byte_count: u64
+        out[32..40].copy_from_slice(&self.byte_count.to_le_bytes());
+
+        // x_buf: [u8; 64]
+        out[40..104].copy_from_slice(&self.x_buf);
+
+        // x_buf_off: usize
+        // in general, a usize should be serialized into a u64, but in this case, it can't ever be larger than 64
```
npajkovsky, line 338: However, usize can be u32 on 32-bit platforms.

crypto/sha2/src/sha256.rs
```diff
@@ -306,3 +303,67 @@ impl<PARAMS: SHA2Params> Hash for SHA256Internal<PARAMS> {
         SecurityStrength::from_bytes(PARAMS::OUTPUT_LEN / 2)
     }
 }
+
+impl<PARAMS: SHA2Params> SerializableState<108> for SHA256Internal<PARAMS> {
+    fn serialize_state(&self) -> [u8; 108] {
+        let mut out_to_return = [0u8; 108];
```
npajkovsky: Where 108 and other sizes coming from?

crypto/sha2/Cargo.toml
```diff
@@ -6,6 +6,7 @@ edition.workspace = true
 [dependencies]
 bouncycastle-core.workspace = true
 bouncycastle-utils.workspace = true
+serde = { version = "1.0.228", features = ["derive"] }
```
npajkovsky: Is it used?

crypto/sha2/src/sha256.rs
ounsworth: The internal state of SHA2. It is called `H^i_j` in FIPS 180-4, see sections 2.2 and 6.2.

crypto/sha2/Cargo.toml
ounsworth: Oops. My IDE added that and I thought I had removed it. Good catch.

crypto/sha2/src/sha256.rs
ounsworth: That's true, but I don't think it affects the comment, right? For example, serde encodes a usize into a u64 because that's the largest it can ever be on any supported architecture. I think. Encoding a u32 into a u64 is perfectly fine.

I'm not sure I understand the point of your comment.

crypto/sha2/Cargo.toml
dghgit: What's an IDE? ;-)

dghgit:
It looks okay, it'll really help with SLH-DSA as storing intermediate state is almost a requirement to getting the algorithm to work in real time... two things, and you may have to pardon my rust if I've got this wrong. It looks like the version check is automatically rejecting anything older, shouldn't that be newer? Ideally you'd be compatible 1 or 2 versions older, newer might be a bit problematic as you might not know the format. Also, for what it's worth, the LLM suggested LIB_VERSION could be derived from env!("CARGO_PKG_VERSION") with the actual version been something taken from the core API crate.

ounsworth:
> It looks like the version check is automatically rejecting anything older, shouldn't that be newer?

My idea was to have a framework to be able to carry multiple versions of the parser in case we ... I don't know ... add a boolean flag to one of the objects for some reason.

I honestly hadn't thought about blocking future versions. That's a good idea!


> Also, for what it's worth, the LLM suggested LIB_VERSION could be derived from env!("CARGO_PKG_VERSION").

I also did some research but didn't find anything useful. Let me give this a bit more poking. Thanks David!

crypto/core/src/serializable_state.rs
ounsworth: Turns out rust auto-implements Ord for byte arrays:

```
assert!([0, 0, 0] < [0, 0, 1]);
```

ounsworth:
I don't know why 2d0b7f3e20290a5db275981b4ba5807f733a9c3c got included in this PR. I also don't know how to remove it. Sorry.

crypto/sha2/src/sha256.rs
npajkovsky: I should have expressed myself in sentences. It seems that you care about endianess with `to_le_bytes()`. If that's true, and you can move a state across endianness, you should also care about moving a state across 32/64 bit CPUs.

If you serialise the state on 32 bits, usize is 4 bytes, and deserialise on 64, you will have to read 8 bytes.

ounsworth: Right. So I think that the answer here is to not let it serialize a `usize` per the local architecture; instead I should always pick a concrete size that will be the same regardless of architecture, right? I think we're on the same page?

In general, that would mean always encoding a `usize` as a `u64` (which I believe is what serde does), but in this particular case, the sha2 `x_buf` is 64 bytes, therefore the `x_buf_off` can never be larger than 64, which actually fits in a u8 ... so in this case there's no need to use 8 bytes when 1 byte will do.

This makes sense, right? I'm not talking crazy?

crypto/sha2/src/sha256.rs
ounsworth: I have replaced all the magic numbers with named pub const's to make it clearer.

crypto/sha2/src/sha256.rs
npajkovsky: > Right. So I think that the answer here is to not let it serialize a usize per the local architecture; instead I should always pick a concrete size that will be the same regardless of architecture, right? I think we're on the same page?

Exactly.

> In general, that would mean always encoding a usize as a u64 (which I believe is what serde does), but in this particular case, the sha2 x_buf is 64 bytes, therefore the x_buf_off can never be larger than 64, which actually fits in a u8 ... so in this case there's no need to use 8 bytes when 1 byte will do.

Yes, that's true.

> This makes sense, right? I'm not talking crazy?

No, that makes sense.

crypto/hkdf/src/lib.rs
```diff
@@ -713,3 +775,105 @@ impl<H: Hash + HashAlgParams + Default> KDF for HKDF<H> {
...
+                state: [u8; $serialized_hkdf_len],
+                salt: &Self::Key,
+            ) -> Result<Self, SerializedStateError> {
+                // Rebuild the salt-keyed HMAC (first in the payload) by re-supplying the salt.
+
+                // This double-dips on HMAC checking the version tag before going any further.
+                // If ever we need to version-reject HKDF separately from HMAC, then we'll need to add
+                // an explicit version check here, and change "None" to the oldest accepted version.
+                // _ = check_lib_ver(&serialized_state, None)?;
+                let hmac = match state[$serialized_hmac_len] {
```
officialfrancismendoza: from_suspended doesn't cross-check the hmac-present flag against the HkdfStates tag. A blob with the present-flag = 0 but state = Initialized/TakingAdditionalInfo deserializes fine and then panics on self.hmac.unwrap() in do_extract_final/do_extract_update_*. This panic is a DoS.

Per the trait contract, this should be rejected with InvalidData. Suggest adding if (hkdf_state == HkdfStates::Uninitialized) != hmac.is_none() { return Err(SerializedStateError::InvalidData) } and a regression test.

crypto/core/src/suspendable_state.rs
```diff
@@ -0,0 +1,111 @@
...
+        // A component > 255 overflows u8 and fails the build (SemVer fields are u8 by design).
+        result = result * 10 + (d - b'0');
+        i += 1;
+    }
+    result
+}
+
+/// The current library version, taken from this crate's `Cargo.toml` at compile time (via Cargo's
+/// `CARGO_PKG_VERSION_*` env vars) so it can never drift from the published version.
+pub const LIB_VERSION: SemVer = SemVer {
```
officialfrancismendoza, line 55: LIB_VERSION comes from core's Cargo.toml, which is 0.1.1 here (release is 0.1.2alpha). 

Stamp doesn't match the release. As the only compat (compatbility) gate, a format-breaking change in a primitive crate (example: sha2, sha3, hkdf, etc.) wouldn't bump core, hence NOT being rejected. 

The old blob would be misplaced. Can we reconcile versions and add a maintainer note that any layout change must bump core's version + set not_before?

crypto/hkdf/src/lib.rs
officialfrancismendoza: > state[$serialized_hmac_len] == 0

The inner HMAC is never parsed, so the version header is never checked. Maybe we should uncomment the check_lib_ver(&state, None)? at the top so version gating holds even when hmac is absent

crypto/sha3/src/keccak.rs
```diff
@@ -343,6 +341,149 @@ impl KeccakDigest {
...
+        // data_queue: [u8; 192]
+        let data_queue: [u8; 192] = input[200..392].try_into().unwrap();
+
+        // bits_in_queue: usize. It can never legitimately exceed the rate (which is at most 168
+        // bytes, well within data_queue's 192-byte capacity).
+        let bits_in_queue = u64::from_le_bytes(input[392..400].try_into().unwrap()) as usize;
+        if bits_in_queue > rate {
+            return Err(SerializedStateError::InvalidData);
+        }
+
```
officialfrancismendoza: from_serialized_state doesn't fully validate bits_in_queue. A blob with squeezing = false and an odd bits_in_queue (or == rate) deserializes OK, then panics on the next absorb/pad_and_switch ("attempt to absorb with odd length queue" / the debug_assert). 

Suggest rejecting those here with InvalidData, so tampered states fail cleanly. Corrupt input should be made as an InvalidData, not a panic. Also dangerous since it has a broader blast radius, as SHA3/SHAKE and ML-DSA (via MuBuilder) suspend depends on it.

crypto/hmac/src/lib.rs
```diff
@@ -420,3 +470,70 @@ impl<HASH: Hash + Default> MAC for HMAC<HASH> {
...
+/// Length in bytes of the serialized state of [HMAC_SHA3_224].
+pub const SUSPENDED_HMAC_SHA3_224_STATE_LEN: usize = SUSPENDED_SHA3_STATE_LEN;
+/// Length in bytes of the serialized state of [HMAC_SHA3_256].
+pub const SUSPENDED_HMAC_SHA3_256_STATE_LEN: usize = SUSPENDED_SHA3_STATE_LEN;
+/// Length in bytes of the serialized state of [HMAC_SHA3_384].
+pub const SUSPENDED_HMAC_SHA3_384_STATE_LEN: usize = SUSPENDED_SHA3_STATE_LEN;
+/// Length in bytes of the serialized state of [HMAC_SHA3_512].
+pub const SUSPENDED_HMAC_SHA3_512_STATE_LEN: usize = SUSPENDED_SHA3_STATE_LEN;
+
+/// HMAC is a keyed algorithm, so it implements [SuspendableKeyed] (rather than
```
officialfrancismendoza, line 522: Should add a security considerations note here: the HMAC/HKDF suspended blob is the inner-hash state over K ⊕ ipad / the secret IKM: specifically, it is derived from secret material. It can't be used to forge a tag without the key, but it does commit to the key and should be stored as confidentially as the key. 

The current wording ("the key is NOT included") reads as "the blob is safe to cache anywhere," which I don't think we want. 

Minor note: Separately: HMAC holds key: [u8; N] but impls neither Secret nor Drop. Preexisted before this PR,, but relevant.)

crypto/core/src/traits.rs
```diff
@@ -470,6 +488,69 @@ pub trait RNG {
...
+/// Allows a stateful object to suspend its operation by serializing its state into a byte array
+///so that it can be resumed later, potentially from a different host.
+///
+/// This is intended for situations where an object is being used through its streaming API
+/// (do_update, do_final) and the operation wants to be paused to a cache, for example while waiting
+/// for network IO.
+///
+/// This is not intended as a mechanism to clone the state of an object since in most cases `.clone()`
+/// will be more straightforward.
+///
```
officialfrancismendoza, line 736: Same concern regarding a security considerations note from earlier

crypto/hkdf/src/lib.rs
```diff
@@ -144,16 +144,63 @@
 //! let mut okm = KeyMaterial::<200>::new();
 //! let _bytes_written = HKDF_SHA256::extract_and_expand_out(&salt, &ikm, info, 200, &mut okm).unwrap();
 //! ```
+//!
```
officialfrancismendoza, line 147: Same concern regarding security considerations note

crypto/core/src/errors.rs
```diff
@@ -72,6 +72,17 @@ pub enum RNGError {
...
+#[derive(Debug)]
+pub enum SerializedStateError {
+    /// The serialized state was produced by a library version incompatible with this one.
+    IncompatibleVersion,
+    /// The serialized state is malformed or corrupt.
+    InvalidData,
+    /// The key supplied to [crate::traits::SuspendableKeyed::from_suspended] does not
+    /// match the key the state was created with (it is bound to a different public-key hash `tr`).
+    IncorrectKey,
+}
```
officialfrancismendoza, line 81: IncorrectKey is never constructed, and its doc claims tr-based wrong-key detection that isn't implemented (MuBuilder::from_suspended takes no key; HMAC/HKDF can't detect it). 

Recommend either wire up the detection or remove the variant ,so callers don't assume wrong-key protection that doesn't exist.

crypto/sha2/src/lib.rs
```diff
@@ -53,8 +88,8 @@ pub const SHA512_NAME: &str = "SHA512";
 /*** pub types ***/
 pub type SHA224 = SHA256Internal<SHA224Params>;
 pub type SHA256 = SHA256Internal<SHA256Params>;
-pub type SHA384 = Sha512Internal<SHA384Params>;
-pub type SHA512 = Sha512Internal<SHA512Params>;
+pub type SHA384 = SHA512Internal<SHA384Params>;
+pub type SHA512 = SHA512Internal<SHA512Params>;
```
officialfrancismendoza, line 92: Source-breaking changes to public items:
> Sha512Internal -> SHA512Internal

Recommend a "breaking changes" entry to release notes here

crypto/mldsa-lowmemory/src/mldsa.rs
```diff
@@ -1210,7 +1287,7 @@ impl<
     /// Internal function to verify a signature 𝜎 for a formatted message 𝑀′ .
     /// Input: Public key 𝑝𝑘 ∈ 𝔹32+32𝑘(bitlen (𝑞−1)−𝑑) and message 𝑀′ ∈ {0, 1}∗ .
     /// Input: Signature 𝜎 ∈ 𝔹𝜆/4+ℓ⋅32⋅(1+bitlen (𝛾1−1))+𝜔+𝑘.
-    fn verify_mu_internal(pk: &PK, mu: &[u8; 64], sig: &[u8; SIG_LEN]) -> bool {
+    fn verify_mu(pk: &PK, mu: &[u8; 64], sig: &[u8; SIG_LEN]) -> Result<(), SignatureError> {
```
officialfrancismendoza, line 1290: Another public breaking change:
> verify_mu_internal -> bool to verify_mu -> Result object

Fine for an alpha, but please add "Breaking Changes" entry to the release notes

crypto/core/src/suspendable_state.rs
```diff
@@ -0,0 +1,111 @@
...
+///
+/// Note that for testability, this will always reject if the serialized state contains a version tag
+/// of `[0,0,0]`.
+///
+/// Hands back a slice to the same array, starting after the version tag.
+pub fn check_lib_ver<const SERIALIZED_LEN: usize>(
+    state: &[u8; SERIALIZED_LEN],
+    not_before: Option<[u8; 3]>,
+) -> Result<&[u8], SerializedStateError> {
+    let ver_bytes: [u8; 3] = state[..3].try_into().unwrap();
```
officialfrancismendoza, line 105: (Nit) Per QUALITY_AND_STYLE, each .unwrap() needs a justifying comment. These slice→array unwraps are infallible by const sizing 

Recommend adding a short // infallible: ... comment on each and confirm quality_stats.sh fallibility didn't regress?

crypto/core/src/serializable_state.rs
```diff
@@ -0,0 +1,111 @@
...
+    if ver < not_before {
+        return Err(SerializedStateError::IncompatibleVersion);
+    };
+    // Nothing is ever compatible with [0,0,0]
+    if ver == SemVer::from([0, 0, 0]) {
+        return Err(SerializedStateError::IncompatibleVersion);
+    };
+
+    // Also not compatible with future versions.
+    if ver > LIB_VERSION {
```
officialfrancismendoza: (Nit) A state produced by a newer build (even a patch bump) is rejected by an older build.  One sentence to specify: states are backward-compatible only (newer lib can read older blob down to not_before; older lib rejects newer blob).

officialfrancismendoza:
Added comments. Ranges from nits, to missing tests + guards that should be added. Additional tests should include:

- **HMAC suspend/resume with an over-length key (>block):** the round-trip test uses a 32-byte key, so the pre-hash branch of load_key_material inside from_suspended is never exercised on the resume path. Recommend adding a 200-byte key case.
- **SHAKE suspend/resume mid-squeeze:** current tests only suspend during absorb. Suspend after squeezing some output, resume, assert the remaining output matches.
- **HMAC wrong-key resume (characterization):** assert that resuming with a different key produces a different tag — locks in the documented "silent" behavior so it can't regress to something worse (e.g. a panic) unnoticed.

crypto/core/src/suspendable_state.rs
ounsworth: Yeah, we certainly have to make sure all the cargo files are bumped to "0.1.2", but that's unrelated to this PR. I'll go do that on the release branch.

crypto/sha3/src/keccak.rs
ounsworth: This one might be better for you to add the guard.
I could probably do it based on this description (or I could feed your description into claude), but if you know exactly which conditions need to to be tested for, maybe better if you do it?

crypto/hmac/src/lib.rs
ounsworth: Yeah, I also thought about adding a security consideration on this point, and decided against it.
My reasoning is this:
I assume it's sorta implied that serializing the internal state of a crypto object ... well, exports the internal state of a crypto object, so you should handle it accordingly. I guess I could add a note to that effect to the top-level Suspendable trait.

In this particular case. You cannot extract the long-term key from the state because that would require you to break the hash function.
As you note, you also can't finish the HMAC / HKDF without the proper key being re-supplied.
So, it's not what "bad thing" an attacker could do if they get their hands on an HMAC / HKDF state, so I would think you actually could cache these on your public github and nothing bad would happen.

(I try to keep security consideration notes to a minimum because if you start putting low-value ones on everything then the really important ones won't jump out as much)

ounsworth: > Minor note: Separately: HMAC holds key: [u8; N] but impls neither Secret nor Drop. Preexisted before this PR,, but relevant.)

That's actually a pretty significant omission! I'll fix that.

Fixed in [07889fc](https://github.com/bcgit/bc-rust/pull/48/commits/07889fcc9cc260bebc1c468929752b3c78ba822f).
That ended up being a more involved fix than I expected because I ran into some rust compiler business when I added Drop to the HMAC struct. Had to go learn how to use core::mem::take. New skill acquired!

crypto/core/src/errors.rs
ounsworth: Good catch. I added that when I was impl'ing Suspendable for ML-DSA, but ended up backing that out and only impl'ing it for ML-DSA::MuBuilder. So that was a leftover.
Now removed.

crypto/sha2/src/lib.rs
ounsworth: The whole library is currently declared ALPHA for exactly this reason: we're still actively changing it and every commit is making braking changes. There's even a big warning box at the top of the main readme saying as much.

We'll start paying more attention to this sort of thing once we declare a 1.0.0 release.

crypto/mldsa-lowmemory/src/mldsa.rs
ounsworth: The entirety of ML-DSA is a new feature since the 0.1.1 release. You can't have a breaking change in a feature that didn't exist before.

Also, we're still in ALPHA exactly for this reason: so that we can just build the library without worrying about breaking changes.

crypto/core/src/suspendable_state.rs
ounsworth: Actually, that points out a potential bug: we should check that `state` is at least 3 bytes long. (in practice this won't be testable from the outside since each object that uses this will static-size their array, but it could use a guard here).
Fixed.

crypto/core/src/serializable_state.rs
ounsworth: This comment got me thinking. Maybe it should accept anything from a newer patch stream as itself? It's literally the definition of `PATCH` that you don't make breaking changes.
https://semver.org/

I'll adjust the docstring :+1:

ounsworth: Fixed in 07889fc

crypto/hkdf/src/lib.rs
ounsworth: This comment is a bit confusing.
I think you mean that HMAC might be absent (since it's optional), in which case it _might not_ be parsed, not that it is _never_ parsed?

But fair point, the possibility of an absent HMAC is a pretty good reason to add those extra 3 bytes so that this always carries a lib_ver tag, independently of the HMAC blob.

Fixed in 3c8d64b

ounsworth: Fixed.

ounsworth:
Merged

PR #52, opened 2026-07-09
forbid(missing_docs) and nightly compiler

ounsworth:
This PR stacks on top of #48 

This PR is going to be a round-up of a bunch of docs changes and refactorings related to removing the nightly compiler requirement.

- [x] Remove dependence on nightly / unstable
- [x] All crates have `#![forbid(missing_docs)]` (and missing docs are added)

As a side-effect of removing nightly / unstable, I refactored how OIDs are carried for HashMLDSA to a more general trait in core, which actually is a much better design anyway.


crypto/base64/src/lib.rs
```diff
@@ -81,6 +81,9 @@
 //     /// "ABCDEFGHIJKLMNOPQRSTUVWXYZabcdefghijklmnopqrstuvwxyz0123456789-_="
 //     URLSafe,
 
+#![forbid(unsafe_code)]
+#![forbid(missing_docs)]
```
npajkovsky, line 85: If you want to maintain your sanity at a reasonable level, try this

root/Cargo.toml

```
[workspace.lints]
rust.missing_docs = "warn"
```

and each crate will have in Cargo.toml

```
[lints]
workspace = true
```

ounsworth:
@officialfrancismendoza  I had to comment out his test at keccak.rs:545 because it's failing and I can't figure out what it's supposed to be testing, so I'll need you to debug it.

crypto/base64/src/lib.rs
ounsworth: I'll give that a try. Thanks.

ounsworth: I tried that with 

```
[workspace.lints]
rust.missing_docs = "forbid"
```

in the workspace-level Cargo.toml, and

```
[lints]
workspace = true
```

in the crate-level Cargo.toml's, and yes that has the same effect, but I'm not a big fan because
A) as you're developing or debugging a crate, you might want to turn off specific lints and then turn them back on later, and 
B) we may not want the same lints on every crates; for example I currently don't have `forbid(missing_docs)` on the `cli` or `mem_usage_benches` crates since those are not exposed as library crates.

So thank you for the suggestion, but I think I'll leave it the way it is.

PR #57, opened 2026-07-13
Complete re-do of Secret and Zeroize

### LINKED ISSUE #32: Apply Secret trait consistently across the library

ounsworth:
The old way we were doing Secret to force a zeroizing Drop was well-intentioned, but likely completely useless since by definition nothing can read a value after it is Drop'd, so the compiler is well within its rights to elide that write. This complete redo of the whole zeroizing architecture does a few things:

1. Concentrate all the Secret behaviours (zeroizing Drop / Debug / Display) into a new module `bouncycastle-utils::Secret`.
2. The new object `Secret<T>` can be wrapped around any `u8, u16, u32, u64, u128, usize, i8, i16, i32, i64, i128, isize, bool, and char` or arrays thereof. It impl's Deref and DerefMut, so its presence is largely transparent.
For example:
    ```
    struct HMAC<..> {
      key: [u8; LEN],
    }
    ```
    becomes:
    ```
    struct HMAC<..> {
      key: Secret<[u8; LEN]>,
    }
    ```
    and everything just works. `Secret<T>` doesn't even add any memory overhead ... ie `size_of(T) == size_of(Secret<T>)`, which is cool.
3. The Secret implementation uses (very small) unsafe blocks to ensure that the zeroizing write (and constant-time comparison) can't be elided or short-circuited by the compiler). This means that the `bouncycastle-utils` crate is no longer `#![forbid(unsafe_code)]`, which is probably right because eventually we'd need that for constant-time behaviour anyway, but with this architecture, the unsafe rust is contained to the utils crate and does not infect the algorithm crates.
4. This means that structs like HMAC, MLDSAPrivateKey, etc no longer need to be responsible for zeroizing Drop or redacting Debug / Display (which, honestly, was fairly error-prone), so we get to delete a bunch of code that we don't need anymore, including the Secret trait that we used to have..

If merged, replaces #36 and closes #32.

Reviewers should start with `crypto/utils/src/secret.rs` as this is the core new thing that needs to be scrutinized.

The rest is a massive refactor to use the new construct (which went surprisingly easily, which I guess validates the design).

This PR is almost a net decrease in lines-of-code since it removes the requirement for all structs to impl Drop, Debug and Display, and instead delegates that to the new `bouncycastle-utils::Secret` object.

### DIFF
--- a/crypto/utils/src/secret.rs
+++ b/crypto/utils/src/secret.rs
@@ -176,7 +176,7 @@ impl<T: ZeroizablePrimitive> Secret<T> {
     /// Populate it afterwards in place via [`DerefMut`] (e.g. by having an RNG or KDF write into
     /// `&mut *secret`), which avoids ever materializing an unprotected copy of the secret.
     #[inline]
-    pub fn new() -> Self {
+    pub const fn new() -> Self {
         Self(T::ZEROED)
     }
 }
--- a/crypto/utils/src/secret.rs
+++ b/crypto/utils/src/secret.rs
@@ -12,6 +12,7 @@
 //! followed by a [compiler_fence] with [SeqCst](Ordering::SeqCst) ordering so the volatile
 //! writes are not reordered with respect to later memory operations.
 
+use crate::ct;
 use core::any::type_name;
 use core::fmt;
 use core::mem::size_of;
@@ -32,14 +33,14 @@ use core::sync::atomic::{Ordering, compiler_fence};
 //           Secret over a type that impls Drop. So we don't actually care about the underlying type
 //           impl'ing Copy; we only care that it doesn't impl Drop, and the `Copy` bound is a
 //           convenient way to catch that if we in the future do impl_zero_init!(T) for a T that has Drop.
-trait ZeroInit: Copy {
+trait ZeroizablePrimitive: Copy {
     /// The zeroed value of this type.
     const ZEROED: Self;
 }
 
 macro_rules! impl_zero_init {
     ($($t:ty),+ $(,)?) => {$(
-        impl ZeroInit for $t {
+        impl ZeroizablePrimitive for $t {
             const ZEROED: Self = 0 as $t;
         }
     )+};
@@ -47,17 +48,17 @@ macro_rules! impl_zero_init {
 
 impl_zero_init!(u8, u16, u32, u64, u128, usize, i8, i16, i32, i64, i128, isize);
 
-impl ZeroInit for bool {
+impl ZeroizablePrimitive for bool {
     const ZEROED: Self = false;
 }
 
-impl ZeroInit for char {
+impl ZeroizablePrimitive for char {
     const ZEROED: Self = '\0';
 }
 
 /// Blanket impl for fixed-size arrays of any length.
 /// This is an alternative to `[T; N]: Default` which is capped at `N <= 32`.
-impl<T: ZeroInit, const N: usize> ZeroInit for [T; N] {
+impl<T: ZeroizablePrimitive, const N: usize> ZeroizablePrimitive for [T; N] {
     const ZEROED: Self = [T::ZEROED; N];
 }
 
@@ -142,6 +143,20 @@ impl<T: ZeroInit, const N: usize> ZeroInit for [T; N] {
 /// assert_eq!(format!("{secret:?}"), "Secret<u32>(<redacted>)");
 /// ```
 ///
+/// # Memory Usage
+///
+/// As a direct wrapper of the type `T`, a `Secret<T>` does not add any memory overhead.
+///
+/// ```
+/// use bouncycastle_utils::Secret;
+///
+/// print!("{}\n", size_of::<i32>());          // 4
+/// print!("{}\n", size_of::<Secret<i32>>()); // also 4
+///
+/// print!("{}\n", size_of::<[u8; 32]>());          // 32
+/// print!("{}\n", size_of::<Secret<[u8; 32]>>());  // also 32
+/// ```
+///
 /// # 🚨 Security 🚨
 ///
 /// What this does NOT guarantee:
@@ -153,9 +168,9 @@ impl<T: ZeroInit, const N: usize> ZeroInit for [T; N] {
 /// * Create a new [Secret] instance, then get a mut ref to its internal value via [Secret::deref_mut]
 ///   and write to that instead of having a copy in an unprotected variable and then copying it into the Secret.
 /// * Avoid copying out of Secret for the same reason.
-pub struct Secret<T: ZeroInit>(T);
+pub struct Secret<T: ZeroizablePrimitive>(T);
 
-impl<T: ZeroInit> Secret<T> {
+impl<T: ZeroizablePrimitive> Secret<T> {
     /// Create a new `Secret` in a zeroed state.
     ///
     /// Populate it afterwards in place via [`DerefMut`] (e.g. by having an RNG or KDF write into
@@ -166,14 +181,14 @@ impl<T: ZeroInit> Secret<T> {
     }
 }
 
-impl<T: ZeroInit> Default for Secret<T> {
+impl<T: ZeroizablePrimitive> Default for Secret<T> {
     #[inline]
     fn default() -> Self {
         Self::new()
     }
 }
 
-impl<T: ZeroInit> Secret<T> {
+impl<T: ZeroizablePrimitive> Secret<T> {
     /// Securely overwrite the contained value with zeros.
     /// After this returns, every byte of the wrapped value has been volatile-written to `0`.
     ///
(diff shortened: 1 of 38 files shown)


crypto/utils/src/secret.rs
```diff
@@ -0,0 +1,286 @@
...
+        let len = size_of::<T>();
+        for i in 0..len {
+            // SAFETY: `base` points at a live, initialized `T`, so the `len` bytes `base..base+len`
+            // all lie within that single allocated object; `i` never reaches `len`, so `add(i)` is
+            // in bounds. `u8` has alignment 1, so every byte address is trivially well-aligned, and
+            // `u8` has no drop glue. `write_volatile` (rather than a plain store) is what forbids
+            // the compiler from eliding this scrub as a dead write. All-zero is a valid bit pattern
+            // for the primitive scalar/array types this wrapper is intended to hold.
+            unsafe {
+                ptr::write_volatile(base_ptr.add(i), 0u8);
```
npajkovsky: The overall implementation is sound and way better than what it was. However, this bit requires a final touch.

Having a loop where every byte written into the index is volatile is too much. You just need to prevent the variable from being elided as a whole, so marking the base is just enough.

You can even simplify it further, and use `core::hint::black_box;` on base_ptr. `black_box` uses inline assembly and marks base_ptr variable as clobbered memory. The trick is that compiler does not understand asm, hence cannot optimize it.

crypto/utils/src/secret.rs
```diff
@@ -0,0 +1,286 @@
...
+        let base_ptr = (&mut self.0 as *mut T).cast::<u8>();
+        let len = size_of::<T>();
+        for i in 0..len {
+            // SAFETY: `base` points at a live, initialized `T`, so the `len` bytes `base..base+len`
+            // all lie within that single allocated object; `i` never reaches `len`, so `add(i)` is
+            // in bounds. `u8` has alignment 1, so every byte address is trivially well-aligned, and
+            // `u8` has no drop glue. `write_volatile` (rather than a plain store) is what forbids
+            // the compiler from eliding this scrub as a dead write. All-zero is a valid bit pattern
+            // for the primitive scalar/array types this wrapper is intended to hold.
+            unsafe {
```
tad-fr: Given this assumption, could the implementation not be simplified to : 
```rust
unsafe {
    ptr::write_volatile(self, core::mem::zeroed());
}
```
Maybe I am missing something but I would've thought that achieves identical results without manually looping.

crypto/utils/src/secret.rs
ounsworth: I'm aware of the `black_box` trick, and I considered this also, but the contract on
https://doc.rust-lang.org/std/hint/fn.black_box.html
very clearly says that it is a "best effort"
whereas the contract on
https://doc.rust-lang.org/std/ptr/fn.write_volatile.html 
is much stronger.

crypto/utils/src/secret.rs
ounsworth: Great suggestion!
I have updated to simplify this.

romen:
I had a look at this earlier this morning, and we had some back and forth in private.
I am posting this to summarize (and cleanup, as I was commenting on the phone :) ), the main points.

---

My impression is that there aren't substantial difference between the new `struct Secret` of this PR, and the [`struct Zeroizing` from the zeroize crate](https://docs.rs/zeroize/1.9.0/zeroize/struct.Zeroizing.html).

At this point, I am not strongly advocating against this PR to stick with the RustCrypto zeroize crate, but I'd like to have a discussion about the tradeoffs that come with the choice of implementing "our own `zeroize`".

The downsides to me seem to be:

- we lose the `#![forbid(unsafe_code)]`
  * good that this was moved from `bc-core` to `bc-utils`, and as you noted maybe that crate would need to use unsafe anyway for some constant-time shenanigans in the future
- we have to study the UB around volatile/non-volatile patterns
  * see note at the bottom
- we have to keep up with rustlang/LLVM semantics/evolution
  * this means we will also have to monitor across different target triplets as they become supported by Rust
  * personally to me this seems to be the most impactful (in terms of maintenance)

On the pro side:

- convenience factor of the `struct Secret` design compatibility with the rest of the `bc-rust` type design.
  * *Personal note: I expect this should improve both readability and maintenance.*
- one less external dependency
  * being as self-contained as possible has been one of the driving directions for the `bc-rust` project since the beginning, so I do appreciate how much this is impactful.


I am not advocating against this PR in favor of the RustCrypto `zeroize` crate, but I believe this discussion needs to happen, the tradeoffs evaluated, and the design decision documented.

Please, fire away your considerations!

----

## Note on `volatile` and Undefined Behavior

IIRC, the volatile pattern in the `zeroize` crate has been refined over time to avoid undefined behavior, because there were subtle nuances to the intuitive way of using volatile for this.
My understanding is that they still worked in practice on available LLVM versions, but were not entirely kosher nor future proof.

This is an excerpt from the [`zeroize` docs](https://github.com/RustCrypto/utils/blob/a8897c9d313bac0abcd86d18e3ffdb42e2208be4/zeroize/src/lib.rs#L144-L158) that should sum-up the current state:

> ## What guarantees does this crate provide?
>
> This crate guarantees the zeroing operation can't be "optimized away" by the compiler, as
> ensured by LLVM's volatile semantics.
>
> Previously there were worries that the approach used by this crate (mixing volatile and
> non-volatile accesses) was undefined behavior due to language contained
> in the documentation for `write_volatile`, however after some discussion
> within the [Unsafe Code Guidelines Working Group], [these remarks have been removed] and the
> specific usage pattern in this crate is considered to be well-defined.
>
> All of that said, there is still potential for microarchitectural attacks
> (ala Spectre/Meltdown) to leak "zeroized" secrets through covert channels.
> This crate makes no guarantees that zeroized values cannot be leaked
> through such channels, as they represent flaws in the underlying hardware.
>
> [Unsafe Code Guidelines Working Group]: https://github.com/rust-lang/unsafe-code-guidelines
> [these remarks have been removed]: https://github.com/rust-lang/rust/pull/60972

romen (commented):
One more suggestion:

I believe our previous attempts at using the `zeroize` crate have been negatively impacted by the lack of "real" examples in its documentation.

The documentation of `Secret` here has a similar problem, it shows what to do for a trivial (but relevant) "secret" array of `u8`.
I think we would benefit from adding an example of a more complex struct, with secret and public fields.
Maybe inspired to the `MLKEMPrivateKey`?

ounsworth:
Thanks @romen. Very valuable discussion.
At the moment I am quite comfortable with the contract on [write_volatile](https://doc.rust-lang.org/std/ptr/fn.write_volatile.html), and I appreciate the work that the rustcrypto team did to make it that way. I'm willing to undertake maintenance of this as the compiler evolves -- I think that we'll need to go there eventually anyway for other aspects of side-channel resistance that we won't be able to just borrow from rustcrypto.

At some point, side-channel protection -- especially the kind where you are trying to protect your process against a malicious host / OS / root process -- is a losing battle. At some point you have to say "best effort" somewhere in your docs. That's not to say that we can't increase our level of effort here in the future, but for today I think this PR is a dramatic increase over what we had before.

ounsworth:
@romen How's this for a toy example of applying this to a more complex struct?
I've added that to the docstring for the Secret<T> module.


```rust
use bouncycastle_utils::secret::{Secret, ZeroizablePrimitive};

/// Holds a system user
#[derive(Clone, Copy)]
struct User {
    userid: i32,
    name: [u8; 64],
}

/// Provide the const ZEROED value for the type.
impl ZeroizablePrimitive for User {
    const ZEROED: Self = Self{ userid: 0i32, name: [0u8; 64] };
}

/// We will tag the admins as Secret<User> to give them extra protections against
/// having their info leaked.
struct AllUsers {
    users: Vec<User>,
    admins: Vec<Secret<User>>,
}
```

ounsworth:
The commit above re-works ML-DSA and ML-KEM to make more surgical use of the new Secret<T> type by being more selective about which components are tagged as secret. This gives about a 5% performance increase over the previous commit.


As an example, the MLDSAPrivateKey object is now defined as:

```rust
pub struct MLDSAPrivateKey<..> {
    rho: [u8; 32],
    K: Secret<[u8; 32]>,
    tr: [u8; 64],
    s1_hat: Secret<Vector<l>>,
    s2_hat: Secret<Vector<k>>,
    t0_hat: Vector<k>,
    seed: Option<KeyMaterial<32>>,
}
```

which clearly indicates to a human reader which values are secret (ie the actual private key components) and which are not (ie part of the public key). This also gives developers more fine-grained control over which data pays the extra performance penalty associated with zeroization-on-drop.

ounsworth:
Merged manually, but with a `--squash`, so might not close manually.

Issue #64, opened 2026-07-27
Implement AES Block Cipher


officialfrancismendoza:
Implement the canonically specified FIPS-197 AES block cipher

officialfrancismendoza:
As per Mike's suggestion, pushing to his feature branch, commits tracked [here](https://github.com/ounsworth/bc-rust/commits/feature/aes/), so we can work on it together by splitting work (ie: me section 5.1, him section 5.2, in FIPS 197 AES)

PR #65, opened 2026-07-29
Feat: Implement AES Block Cipher 

officialfrancismendoza:
Implement FIPS 197, 2023 update of AES block cipher (#64)


dghgit:
The final round for this is now on #115

Issue #66, opened 2026-08-05
Factory: refactor into macros?


ounsworth:
Currently the factory objects look like this:

```
pub enum HashFactory {
    ///
    SHA224(sha2::SHA224),
    ///
    SHA256(sha2::SHA256),
    ///
    SHA384(sha2::SHA384),
    ///
    SHA512(sha2::SHA512),
    ///
    SHA3_224(sha3::SHA3_224),
    ///
    SHA3_256(sha3::SHA3_256),
    ///
    SHA3_384(sha3::SHA3_384),
    ///
    SHA3_512(sha3::SHA3_512),
}
...
    fn hash(self, data: &[u8]) -> Vec<u8> {
        match self {
            Self::SHA224(h) => h.hash(data),
            Self::SHA256(h) => h.hash(data),
            Self::SHA384(h) => h.hash(data),
            Self::SHA512(h) => h.hash(data),
            Self::SHA3_224(h) => h.hash(data),
            Self::SHA3_256(h) => h.hash(data),
            Self::SHA3_384(h) => h.hash(data),
            Self::SHA3_512(h) => h.hash(data),
        }
    }
```

This code is highly repetitive and begging for a macro, maybe of the form:

```
impl_hash_factory!(SHA224, SHA256, SHA384, SHA512, SHA3_224, SHA3_256, SHA3_384, SHA3_512)
```

That said, I generally find macros for the sake of reducing the number of lines of code are not worth it because they make the code harder to read, harder to debug, and harder to test. So I think we should do the refactor of bouncycastle-factory to use macros, just to see what it looks like, and then decide if it is actually an improvement or not.

npajkovsky:
I have a better idea. Stay tuned.

PR #76, opened 2026-08-13
Replace Factory boilerplate with derive macros

npajkovsky:
HashFactory hand-wrote an 8-arm match for each of the 10 Hash trait methods plus the string-name lookup in AlgorithmFactory::new, so adding an algorithm meant touching a dozen places. Add a bouncycastle-factory-macros crate providing two derives that generate all of it:

  * Hash             - the Hash impl, forwarding every method to the
                       variant currently held.
  * AlgorithmFactory - Default and AlgorithmFactory (new, default_128_bit,
                       default_256_bit), driven by a #[factory(name = ...)]
                       helper attribute on each variant.

Adding an algorithm is now one variant plus one attribute. hash_factory.rs drops from 254 to 79 lines with no behaviour change; existing tests and doctests cover the generated code.

Each derive is named after the trait it implements - derives live in the macro namespace and traits in the type namespace, so they don't collide (same convention as serde::Serialize).

Also marks HashFactory #[non_exhaustive] so future variants are additive rather than breaking for downstream matches

Fixes: https://github.com/bcgit/bc-rust/issues/66

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -13,6 +13,7 @@ bouncycastle-base64 = { path = "./crypto/base64" }
 bouncycastle-core = { path = "crypto/core" }
 bouncycastle-core-test-framework = { path = "./crypto/core-test-framework" }
 bouncycastle-factory = { path = "./crypto/factory" }
+bouncycastle-factory-macros = { path = "./crypto/factory-macros" }
 bouncycastle-hex = { path = "./crypto/hex" }
 bouncycastle-hkdf = { path = "./crypto/hkdf" }
 bouncycastle-hmac = { path = "./crypto/hmac" }
--- a/crypto/factory-macros/Cargo.toml
+++ b/crypto/factory-macros/Cargo.toml
@@ -1,0 +1,12 @@
+[package]
+name = "bouncycastle-factory-macros"
+version.workspace = true
+edition.workspace = true
+
+[lib]
+proc-macro = true
+
+[dependencies]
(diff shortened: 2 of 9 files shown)


ounsworth:
I didn't know about the `#[non_exhaustive]` attribute. That's great. Regardless of whether we macro the factories or not, we should apply that, and I will especially be adding `#[non_exhaustive]` to all the error type enums.

I particularly like that it doesn't prevent you, inside the crate, from writing exhaustive matches
for example:

```
        match self {
            Self::SHAKE128(h) => h.hash_xof_out(data, output),
            Self::SHAKE256(h) => h.hash_xof_out(data, output),
        }
```
which will break and force you to add a new branch if you add a new type to the struct, but any caller _outside the library_ would be required to add the `_ =` branch if they tried to do the same thing. Brilliant! Thanks for pointing that out.

I created: #79

npajkovsky:
@jjkurczak @ounsworth I have pushed the KDF factory proc-macro.

dghgit:
I found this one quite interesting. I think we should just park the factory stuff for now though - where it's going to become important is when we start adding PKIX support, but if the example of bc-java/bc-csharp is anything to go by it's going to be based on the use AlgorithmIdentifier structures and not Strings. So I think the factory approaches, while going on the bench now, will get their day in the Sun, we'll just wait for when we have a clearer picture of what we're feeding into them.

npajkovsky:
@dghgit I'll park it in my local branch for now.

Issue #71, opened 2026-08-10
Ensure that SHA2 and SHA3 crates are #![no_std]


jjkurczak:
### Summary:
The purpose of this sub-issue is to implement sufficient no_std changes such that SHA2 and SHA3 crates are able to be built, tested, and benchmarked in no_std.

### Scope:
The expected changes for this initial task may involve changes to other crates, and the workspace, to support required changes to crate dependencies, and to create a baseline for all crates to implement the new functionality uniformly.

### Acceptance Criteria:
- [ ] A new feature will be added that positively enables std, which will be on by default
- [ ] the `--no-default-features` flag will be used to enforce no_std at build time
- [ ] SHA2 and SHA3 crates must build/test/bench, individually, with `--no-default-features`
- [ ] All crates must continue to build/test/bench, with no regressions, in default config.
- [ ] Building other crates or the entire workspace with `--no-default-features` is unsupported, with no required behavior associated with acceptance of this sub-issue

PR #77, opened 2026-08-19
Ensure that SHA2 and SHA3 crates are #![no_std]

jjkurczak:
## Issue Link
resolves bcgit/bc-rust#71

## Summary
Updates SHA2 and SHA3 packages to build with no_std, and configures workspace and other packages to support no_std

### What did you do?
Beginning with preliminary work from @officialfrancismendoza to support building with no_std, I updated the workspace and all crypto package Cargo.toml files to consistently add an `alloc` feature to each package, which is enabled by default, and configured the workspace to disable default features for all packages when included as a dependency within the workspace.
I also added `#![cfg_attr(not(feature = "alloc"), no_std)]` to the relevant crate root lib.rs files to actually disable std according to the `alloc` feature for that lib.

I updated `core/source/traits.rs` to block all std-dependent functions from being included in trait definitions when not using the `alloc` feature, which has a significant impact on all the remaining packages when building without `alloc`.
A majority of the changes in this PR result from needing to disable code attempting to use these blocked alloc-only trait functions, include many uses in tests/ directories. Note that, as a result, there are tests which have been disabled in order to get this change to compile and run. All tests that do not appear to have a no_std equivalent for validating behavior have been marked with `// todo: may require no_std equivalent`.

I then built and ran tests for the sha2 and sha3 packages, incrementally fixing anything that was breaking due to missing std functionality.
This required changes to their core-test-framework and hex dependencies to be able to build under no_std.

Finally, I added a `no_std_utils.rs` module to `core-test-framework` to contain a Vec-returning `hex_decode` function that is missing from the hex crate when building with no_std.
Hex decoding of strings into Vecs is a common pattern used with many tests, so this utility lets most of those uses remain with only a trivial find/replace to use the no_std_utils version.
Integration tests are unique targets which do not inherit no_std attributes from the crate root, so it should safe to add further helpful std-dependent utilities to `no_std_utils.rs` as long as they're used exclusively in integration test files.
Note that this approach means that the integration tests themselves cannot be built with no_std to run in an embedded environment.

### What alternatives did you consider?
There were some choices made in this PR:
1. I decided to disable tests incompatible with no_std rather than fix them, unless they were trivial. There are enough tests requiring attention that it seemed better to break out that work into another task, if the goal is to get the ball rolling on no_std. I am open to hearing reviewers' sense of the priority of rebuilding the missing tests vs continuing to get other crates working with no_std.
1. I added no_std_utils, rather than modify the uses of `hex::decode()` to use `hex::decode_out()` instead - these changes become non-trivial (without AI and careful review) when addressing the hundreds of uses in the remaining packages, so it seemed better to establish a directly-compatible alternative and a utils module to build on for other similar functions in the future.

### Did you add or update any tests to account for these changes?
No tests added - many tests disabled for no_std builds, see reasoning above.

### Is there anything else the reviewer should keep in mind?
1. For the moment, only sha2 and sha3 can build, test, or benchmark individually with --no-default-features.
Building with --no-default-features, or otherwise disabling `alloc`, for any package other than sha2 or sha3 (including when building --workspace), is currently an *unsupported configuration*.
1. No changes have been made to github actions to validate --no-default-features builds - it seems like this should be managed once the entire workspace of packages is capable of being validated for --no-default-features? Is it worth spending the effort to mess around with the github actions for each sub-issue, so each set of packages is validated as they begin to support no_std, and then refactor for the entire workspace functioning at the end?

### Which packages are impacted?
This change impacts most packages. Nearly all changes relate to disabling code when the `alloc` feature is not enabled, so default builds using `alloc` should ideally see no change.

### What runtime behaviors could be impacted?
 It's unlikely that any changes in this PR will impact runtime behaviours - I would expect to see breaks at compile time rather than at runtime.

### What could go wrong in the worst-possible scenario?
There are two major areas of risk I've identified:
1. There are widespread changes to workspace and package .toml files to create the `alloc` feature in most packages, and to ensure that dependencies only inherit `alloc` when needed. There could be unexpected dependency breakages when bulding packages if the dependencies have been misconfigured.
1. Many tests are currently disabled in --no-default-features for sha2 and sha3. There is a risk that core functionality to these packages could be modified by another PR before these tests are re-enabled, and bugs will be introduced into those packages due to missed failures.

## Validation
1. build, test, bench workspace should continue to function as expected (bench manually)
1. build, test, bench individual packages should continue to work as expected (build, test, bench manually)
1. build, test, bench sha2 and sha3 packages individually with --no-default-features flag should work, with reduced test coverage (test manually)

## AI Usage Statement
- [ ] No
- [X] Yes, trivial code changes were generated by AI (e.g., autocompletion of a single line, reformatting, or spell-checking)
- [ ] Yes, non-trivial code changes were generated by AI

Assisted-by: ClaudeCode:Sonnet5


npajkovsky:
@jjkurczak is it ready for review?

jjkurczak:
@npajkovsky thanks for checking! We're going a different direction for no_std, so I'm closing this PR

PR #81, opened 2026-08-21
Added the `#[non_exhaustive]` attribute

### LINKED ISSUE #79: Make use of the `#[non_exhaustive]` attribute
# Background

Rust's `#[non_exhaustive]` attribute is used to mark structs and enums that are expected to grow over time so that downstream dependent code is not allowed to exhaustively match the types, but instead must always have a `_ =` branch in order to gracefully handle future additions.

For example, consider:

```rust
#[non_exhaustive]
pub enum XOFFactory {
    ///
    SHAKE128(sha3::SHAKE128),
    ///
    SHAKE256(sha3::SHAKE256),
}
```

Within the same crate, we are allowed to do an exhaustive match, which is great because if we add a new type of XOF in the future, we want the compiler to tell us all the places that we need to add a branch. But if an external user of this crate tries to write the same code, the `#[non_exhaustive]` attribute will force them to add a `_ =` arm to future-proof it against new versions of the lib that add a new type.

```
    fn hash_xof_out(self, data: &[u8], output: &mut [u8]) -> usize {
        output.fill(0);

        match self {
            Self::SHAKE128(h) => h.hash_xof_out(data, output),
            Self::SHAKE256(h) => h.hash_xof_out(data, output),
        }
    }
```

# Task

The task is to apply the `#[non_exhaustive]` attribute to any pub enums that are expected to grow over time. Obvious candidates include the factories and error type. Maybe after doing those, ask Claude if there are any other good candidates.

ounsworth:
Added the `#[non_exhaustive]` attribute to enums that are likely to grow in the future.

Closes #79

### DIFF
--- a/crypto/core-test-framework/src/mac.rs
+++ b/crypto/core-test-framework/src/mac.rs
@@ -116,6 +116,10 @@ impl TestFrameworkMAC {
                     low_security_key.set_key_len(64).unwrap(); // truncate should be infallible
                     low_security_key.set_security_strength(SecurityStrength::_192bit).unwrap();
                 }
+                // `SecurityStrength` is `#[non_exhaustive]`, so this arm is required.
+                _ => panic!(
+                    "unhandled SecurityStrength variant -- add a case for it in the MAC test framework"
+                ),
             };
             Ok(())
         })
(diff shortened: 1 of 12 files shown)


npajkovsky (approved):

ounsworth:
Merged manually.

PR #83, opened 2026-08-21
Cosmetic refactoring to MLDSA / MLKEM Polynomial and Matrix

ounsworth:
Polynomial was left pub (and exported outside the crate) as a consequence of the old way of handling secrets, but does not need to be anymore. And some other cosmetic refactoring.


ounsworth:
Merged manually as 57998ee4ae178983b07754b3ddbe4d581c437fa0

Issue #75, opened 2026-08-12
Add a CT validation framework based on valgrind timecop


ounsworth:
Search for "valgrind contant time".

Todo fill in this ticket

Here is one paper:
https://eprint.iacr.org/2024/1049.pdf

npajkovsky:
I'll fill it. I have some work already done.

PR #87, opened 2026-08-26
Feature/sha3 final patches

dghgit:
changes two things:
                                           
  - partial-bit squeeze returned the wrong bits, it was tested, but it appears the test was using 0xff which hid the error.
  - validates  num_partial_bits in do_final_partial_bits so it has to be between 0 and 7. Previously 8 to 15 meant the data was just absorbed and 16 and greater caused a panic.
  
Just to note, I'm getting told that main won't pass the format check... the diffs have been checked though.


crypto/sha3/src/lib.rs
```diff
@@ -108,6 +108,37 @@
...
+//! footprints. No heap memory is used by the algorithms themselves; the `Vec<u8>`-returning
+//! convenience methods allocate only the output buffer, and the `*_out` variants allocate nothing.
+//!
+//! | Object                                  | Size (bytes) |
+//! |-----------------------------------------|--------------|
+//! | `SHA3_224` .. `SHA3_512`, `SHAKE128/256` | 440          |
+//! | Suspended state ([`Suspendable`])       | 415          |
+//!
+//! The object size is dominated by the 200-byte sponge state and a 192-byte input/output queue.
+//! The Keccak permutation itself runs on 25 `u64` locals plus a handful of temporaries, so transient
```
ounsworth: To me, these two lines are explaining internal detail and do not add anything for a user. I suggest deleting them.

crypto/sha3/src/lib.rs
```diff
@@ -108,6 +108,37 @@
...
+//! stack usage per call is on the order of a few hundred bytes beyond the object.
+//!
+//! # Security Considerations
+//!
+//! * SHA3-224/256/384/512 offer 112/128/192/256 bits of collision resistance respectively; SHAKE128
+//!   and SHAKE256 offer 128 and 256 bits of security for output lengths at least twice that size
+//!   (FIPS 202 Appendix A.1).
+//! * SHAKE is an XOF, not a hash: `SHAKE128(m, 32)` is a prefix of `SHAKE128(m, 64)`. If the output
+//!   length must be bound to the digest, include it in the message (FIPS 202 Appendix A.2). For this
+//!   reason SHAKE does not implement [`Hash`].
```
ounsworth: I think that "For this reason SHAKE does not implement [`Hash`]." is a true statement, but not really a security consideration. I suggest deleting it.

crypto/sha3/src/lib.rs
```diff
@@ -108,6 +108,37 @@
...
+//! * Once squeezing begins, no further input can be absorbed; [`XOF::absorb`] returns
+//!   [`HashError::InvalidState`] rather than silently producing an unapproved construction.
```
ounsworth: This also doesn't feel like a security consideration?

Also, I would like opinions on this because some people have pushed back on me that while FIPS 202 clearly illustrates the sponge construction as

    absorb() -> absorb() -> squeeze() -> squeeze()

it doesn't technically prohibit

    absorb() -> squeeze() -> absorb() -> squeeze()

and so bc-rust should relax this requirement. Unrelated to whether this is a "security consideration" or not, I would love your input on this.

crypto/sha3/src/lib.rs
```diff
@@ -108,6 +108,37 @@
...
+//! * The sponge state and queue are held in [`bouncycastle_utils::secret::Secret`] and zeroized on
+//!   drop. Transient copies in registers/stack locals during the permutation are not zeroized.
```
ounsworth: Is this too much internal detail for docs? I don't think we go into this much analysis for any other algs. If we should, then we should do them all in one pass a self-contained task and make them cohesive, not do it ad-hoc.

crypto/sha3/src/lib.rs
```diff
@@ -108,6 +108,37 @@
...
+//! * The implementation contains no data-dependent branches or table lookups.
```
ounsworth: This is stated as a fact, but I think at the moment it's conjecture. We have #75 to build a ct testing framework to validate this.

I feel uncomfortable making ct claims that we haven't validated.

crypto/sha3/src/lib.rs
```diff
@@ -108,6 +108,37 @@
...
+//! * The [`KDF`] entropy tracking is a developer aid, not a security proof: it can only reason about
+//!   the metadata attached to [`KeyMaterial`] inputs.
```
ounsworth: This sentence is going to be true of any function that does any enforcement based on the KeyMaterial metadata, of which there are many across the library. If we really feel the need to say this, then I think we should say it once on the KeyMaterial object, and not on all the objects that consume KeyMaterial.

crypto/sha3/src/lib.rs
```diff
@@ -205,11 +236,13 @@ trait SHA3Params: HashAlgParams {
...
 
-impl HashAlgParams for SHA3_224 {
-    const OUTPUT_LEN: usize = 28;
-    // const BLOCK_LEN: usize = 64;
-    const BLOCK_LEN: usize = 144; // FIPS 202 Table 3
+/// The public hash types expose the same parameters as their `*Params` marker, so the constants
+/// are defined exactly once (on the params struct) and forwarded here.
+impl<PARAMS: SHA3Params> HashAlgParams for SHA3Internal<PARAMS> {
+    const OUTPUT_LEN: usize = PARAMS::OUTPUT_LEN;
+    const BLOCK_LEN: usize = PARAMS::BLOCK_LEN;
```
ounsworth, line 237: Oh sweet! That's an excellent use of generics!! 👍

crypto/sha3/src/sha3.rs
```diff
@@ -150,10 +187,9 @@ impl<PARAMS: SHA3Params> Hash for SHA3Internal<PARAMS> {
         output
     }
 
-    fn hash_out(self, data: &[u8], mut output: &mut [u8]) -> usize {
-        output.fill(0);
-
-        self.hash_internal(data, &mut output)
+    fn hash_out(self, data: &[u8], output: &mut [u8]) -> usize {
+        // hash_internal zeroizes `output` before writing.
```
ounsworth:
```suggestion
```

Good catch to remove the duplicate .fill(0), but the comment is unnecessary.

crypto/sha3/src/sha3.rs
```diff
@@ -44,6 +44,43 @@ impl<PARAMS: SHA3Params> SHA3Internal<PARAMS> {
...
+    /// Appends the SHA3 domain-separation suffix and pads as per FIPS 202 s. 6.1, then squeezes the digest.
+    ///
+    /// `num_partial_bits` (0..=7, validated by the caller) trailing message bits are taken from the
+    /// least significant bits of `partial_byte` (FIPS 202 Appendix B.1 bit ordering). FIPS 202 s. 6.1
+    /// defines SHA3-d(M) = KECCAK[c](M || 01, d), so the two suffix bits are appended directly above
+    /// the message bits; pad10*1 is then applied by the sponge when it switches to squeezing.
+    ///
+    /// Returns the number of bytes written (`min(output.len(), OUTPUT_LEN)`); a shorter output buffer
+    /// truncates the digest, a longer one is zero-filled past the digest.
+    fn finalize(mut self, partial_byte: u8, num_partial_bits: usize, output: &mut [u8]) -> usize {
```
ounsworth: I think you're gonna need to explain this one because I don't understand why you're adding this.

This file already has functions `do_final_partial_bits()` and `do_final_partial_bits_out()`, which is impl'd from core::traits::Hash. This looks like a duplicate?

Yeah. Can you please explain why you've added this?

(if we are going to keep this, the name `finalize()` does not match the naming conventions of the library. Since it consumes self, it should be `do_final`, and since it takes an output buffer, it should be `_out`, ... so it should be called `do_final_<something>_out()`.)

crypto/sha3/src/sha3.rs
```diff
@@ -169,63 +205,35 @@ impl<PARAMS: SHA3Params> Hash for SHA3Internal<PARAMS> {
...
-            self.keccak.squeeze(output)
-        } else {
-            let min =
-                if output.len() >= self.output_len() { self.output_len() } else { output.len() };
-            self.keccak.squeeze(&mut output[..min])
-        };
-        bytes_written
+    fn do_final_out(self, output: &mut [u8]) -> usize {
+        // A whole-byte message is the zero-partial-bits case of the general finalization.
+        self.finalize(0, 0, output)
```
ounsworth: Why is this chaining to a brand-new function, and not to the existing `do_final_partial_bits_out()`?

crypto/sha3/src/sha3.rs
```diff
@@ -169,63 +205,35 @@ impl<PARAMS: SHA3Params> Hash for SHA3Internal<PARAMS> {
...
+        // and 8..=15 would silently absorb garbage. 0 is allowed and is equivalent to do_final_out().
+        if num_partial_bits > 7 {
+            return Err(HashError::InvalidLength("num_partial_bits must be in the range [0,7]"));
         }
-
-        self.keccak.absorb_bits(final_input as u8, final_bits)?;
-
-        let min = if output.len() >= self.output_len() { self.output_len() } else { output.len() };
-        Ok(self.keccak.squeeze(&mut output[..min]))
+        Ok(self.finalize(partial_byte, num_partial_bits, output))
```
ounsworth: Can you please explain this one to me. It's not immediately clear to me why this is an equivalent substitution.

crypto/sha3/src/shake.rs
```diff
@@ -304,10 +302,12 @@ impl<PARAMS: SHAKEParams> XOF for SHAKEInternal<PARAMS> {
         if self.keccak.squeezing {
             return Err(HashError::InvalidState("cannot absorb after squeezing has begun"));
         }
-        if !(1..=7).contains(&num_partial_bits) {
-            return Err(HashError::InvalidLength("must be in the range [0,7]"));
+        // Validate before shifting: `1 << num_partial_bits` on a u16 would overflow for values >= 16,
+        // and 8..=15 would silently absorb garbage. 0 is allowed and simply finalizes with no partial byte.
```
ounsworth: I find this comment more confusing than helpful.
What is the ">= 16" referring to? Because the condition we're checking is "> 7"?


This function is called `absorb_last_partial_byte`, so clearly trying to absorb more than 7 is no longer a "partial byte" and you should be using a different API.
I don't know what this comment is trying to tell me, but it's not that.

crypto/sha3/src/shake.rs
```diff
@@ -355,14 +355,20 @@ impl<PARAMS: SHAKEParams> XOF for SHAKEInternal<PARAMS> {
         output: &mut u8,
     ) -> Result<(), HashError> {
         if !(1..=7).contains(&num_bits) {
-            return Err(HashError::InvalidLength("must be in the range [0,7]"));
+            return Err(HashError::InvalidLength("num_bits must be in the range [1,7]"));
```
ounsworth, line 357: This is a structural change that affects the API contract.
What does FIPS 202 say about whether num_bits=0 is a valid partial byte?
Whatever we do here should be consistent with `absorb_last_partial_byte()` above.

crypto/sha3/src/shake.rs
```diff
@@ -355,14 +355,20 @@ impl<PARAMS: SHAKEParams> XOF for SHAKEInternal<PARAMS> {
...
+        // first squeeze the SHAKE "1111" domain-separation suffix (FIPS 202 s. 6.2) is applied.
+        // Calling keccak.squeeze() here would produce raw Keccak output instead of SHAKE output.
         let mut buf = [0u8; 1];
-        self.keccak.squeeze(&mut buf);
-        *output = buf[0] >> 8 - num_bits;
+        self.squeeze_out(&mut buf);
+
+        // FIPS 202 Appendix B.1 (h2b): the first `num_bits` bits of an output byte are its least
+        // significant bits. This matches the input-side convention used by absorb_last_partial_byte().
+        *output = buf[0] & ((1u8 << num_bits) - 1);
```
ounsworth, line 366: Ah good catch. That seems like a legitimate bug 👍 

I don't think I have done `bc-test-data` or wycheproof tests yet for the sha2 and sha3 crates. Presumably this would have been caught.

That said, I'm not sure that the comment is helpful since this is already stated on the new docstring comment on core::traits::Hash, right? I think this comment should be deleted.

crypto/sha3/src/shake.rs
```diff
@@ -355,14 +355,20 @@ impl<PARAMS: SHAKEParams> XOF for SHAKEInternal<PARAMS> {
...
 
         *output = 0;
 
+        // Route through squeeze_out() rather than keccak.squeeze() directly, so that if this is the
+        // first squeeze the SHAKE "1111" domain-separation suffix (FIPS 202 s. 6.2) is applied.
+        // Calling keccak.squeeze() here would produce raw Keccak output instead of SHAKE output.
         let mut buf = [0u8; 1];
-        self.keccak.squeeze(&mut buf);
-        *output = buf[0] >> 8 - num_bits;
+        self.squeeze_out(&mut buf);
```
ounsworth, line 364: This is changing behaviour and implying that we had a bug before. Great!

I don't think I have done bc-test-data or wycheproof tests yet for the sha2 and sha3 crates. Presumably this would have been caught.

(also, while this comment is helpful to explain the diff, I think we should delete the comment before merging)

crypto/sha3/src/shake.rs
ounsworth: Also, doesn't this change behaviour? The old logic was "if !(1..=7), and the new behaviour is "if > 7". But what about 0?
I think we need to have a careful look at the API docstring and the FIPS doc and see if num_bits = 0 is a valid "partial byte" or not.
(it would make sense to me that it is, because ... why not ... but we should check. And presumably if we do it wrong, then wycheproof and bc-test-data will catch it, once we get around to wiring those up).

(either way, we should align the error message to the final choice)

ounsworth (changes requested):
This is changing quite a bit of code that had been tested with cargo mutants. Have you redone the cargo mutants testing? I can walk you through how to do that.

crypto/sha3/src/lib.rs
```diff
@@ -108,6 +108,37 @@
...
+//! # Memory Usage
+//!
+//! All SHA3 and SHAKE variants share the same Keccak-f\[1600\] sponge and so have identical memory
+//! footprints. No heap memory is used by the algorithms themselves; the `Vec<u8>`-returning
+//! convenience methods allocate only the output buffer, and the `*_out` variants allocate nothing.
+//!
+//! | Object                                  | Size (bytes) |
+//! |-----------------------------------------|--------------|
+//! | `SHA3_224` .. `SHA3_512`, `SHAKE128/256` | 440          |
+//! | Suspended state ([`Suspendable`])       | 415          |
```
ounsworth, line 120: For my curiosity, where did these numbers come from?
Where I've done this on other crates, I have, for example, a `/mem_usage_benches/bench_mldsa_mem_usage.rs` that prints struct sizes using rust's `size_of::<>` operator.

This PR has not added equivalent mem benches for SHA3, so how did you measure these numbers?

ounsworth (changes requested):

ounsworth:
We should probably port the bc-test-data and wycheproof test harnesses over from the mldsa crate as part of this PR.

dghgit:
I've tried to answer stuff inline (should start appearing in a minute), but one note, with the comments, as the branch is also been worked on by an LLM, as a general rule work like this (while on the branch) will likely include comments that may seem unnecessary or out of place. I've learned to leave this alone until the work on the branch is finished as they also form part of the LLM writing notes to itself, so deleting them early is really only useful if you want to increase the chance of an error or something being missed. On merging though, feel free to make whatever edits you want if you feel like something doesn't match the house rules (as it happens LLMs like that sort of consistency as well, just not having it introduced midway through development).

Cargo mutant was redone as well.

One other note: cargo fmt kept trying to change key_material.rs in crypto/core/src, it might need a second look at.

crypto/sha3/src/lib.rs
dghgit: It's a confusion between what is in https://eprint.iacr.org/2011/499 (the original SIG SAC paper) and what's in FIPS PUB 202. Understandable, but FIPS PUB 202 Section 6.2 defines `SHAKE128(M, d) = KECCAK[256](M || 1111, d)` so in the FIPS PUB the construction is actually `absorb() -> pad() -> squeeze()` there is no room for another `absorb()` after the pad. So yes, you might be able to argue that Keccak allows for this, but the SHAKE construction given in FIPS PUB 202 most definitely does not. 

That said, while I don't know what the byte string might look like, I can say `absorb() -> squeeze() -> absorb() -> squeeze()` would also produce a CVE under CWE-682, possibly even CWE-1240 if the reporter was feeling nasty, I guess it would be a medium, not a critical, but even then I doubt any of us would enjoy the grief.

crypto/sha3/src/lib.rs
dghgit: Deleted.

crypto/sha3/src/lib.rs
dghgit: Deleted

crypto/sha3/src/lib.rs
dghgit: It's been moved into the XOF usage docs. If it still seems a bit much, I wouldn't abbreviate it further until this is ready to merge (see comment at the end).

crypto/sha3/src/lib.rs
dghgit: Honestly, I cannot take the full credit for that one.

crypto/sha3/src/lib.rs
dghgit: this has been edited down as well (although see final comment).

crypto/sha3/src/lib.rs
dghgit: Deleted.

crypto/sha3/src/sha3.rs
dghgit: It's been renamed `do_final_bits_out()` it's shared by the other 2 do_final_* methods.

crypto/sha3/src/sha3.rs
dghgit: There's no need for the extra check in do_final_partial_bits_out.

crypto/sha3/src/sha3.rs
dghgit: Following rational was given:

Three things changed, none affecting behaviour:

1. `self.output_len()` → `PARAMS::OUTPUT_LEN`. `output_len()` is defined as `PARAMS::OUTPUT_LEN`
   (see the `Hash` impl a few lines above), so this is the same constant, just resolved at compile
   time rather than via a method call — consistent with how the rest of the impl uses `PARAMS::`.
2. `output.as_mut_slice()` → `&mut output`. `Vec<u8>` deref-coerces to `&mut [u8]`; identical.
3. Removed `dbg_rslt_len` / `debug_assert_eq!(bytes_written, dbg_rslt_len)`. The buffer is allocated
   with exactly `OUTPUT_LEN` bytes and `do_final_partial_bits_out` returns
   `min(output.len(), OUTPUT_LEN)` = `OUTPUT_LEN`, so the assert was checking a tautology about a
   buffer we just sized ourselves.
One thing, the corresponding do_final() still has its assert as I asked it to keep the changes focused. The assert could either be deleted there for consistency, or re-introduced back here for consistency. I'd delete the other as it appears it's also checking something that's now turned into a tautology (I suspect this wasn't always true, but as the code's evolved this has happened).

crypto/sha3/src/shake.rs
dghgit: The comment was pointing out that the old check was needed before `(1 << num_partial_bits)` on a `u16` (which panics in debug / wraps in release for `>= 16`), but "must be 0..=7 because it's a partial byte" is the real justification. I've replaced it, although a little wary about this one as it also feels like memory - I think it's safe though, nothing is likely to role it back by accident.

It is a behaviour change as there's no longer a panic and previously SHAKE::absorb_last_partial_byte rejected 0 with the message "must be in the range [0,7]" which seemed a little odd to me.

crypto/sha3/src/sha3.rs
dghgit: Deleted.

crypto/sha3/src/shake.rs
dghgit: I'm not sure I understand - it's the error message that got changed, it's now correct for the code. The code was always enforcing 1->7

crypto/sha3/src/shake.rs
dghgit: Yes, the bug part 1.

We don't have partial byte tests, I'd be surprised if Wycheproof do either. We can get some from the ACVP though, they will generate them for both SHA3 and SHA2. To be honest, I've never seen partial bytes used, but given the strange world of small devices, I wouldn't be surprised if we'll run into them here... I'll try and get ACVP vectors into bc-test-data for the partial results.

crypto/sha3/src/shake.rs
dghgit: See above. It was tested, but the string was 0xff, so it constituted a fully populated byte. As for the comment, see the comment at the end, I'd do a "house comment" pass before merging as the comments in the branch currently represent a mixture of "comment" and "memory".

crypto/sha3/src/lib.rs
dghgit: As you guessed, it was a throw away though, as I was just focusing on the review issues. Did you want me to commit in something equivalent (would probably suggest a separate branch but can do here).

dghgit:
I've ported the test data harness and it's now in this PR as well. Wycheproof doesn't appear to provide vectors for SHA-3/SHAKE they point people at the NIST ones. I'll update the references in Rust to point at the partial byte vectors when we manage to generate them (it's going to mean tweaking a bit of JSON).

dghgit:
Latest commit adds testing for partial-byte vectors. NIST standard vectors have been added to bc-test-data/crypto/sha3

Note: latest commit also fixes a bug in Keccak concerning partial-bytes. Seems there was one last one lurking in the woods.

crypto/sha3/src/lib.rs
dghgit: I've added mem_usage_benches/bench_sha3_mem_usage.rs as well.

ounsworth:
Replaced by #91 , #92 , #93, #94

PR #88, opened 2026-08-26
Feature/sha2 final patches

dghgit:
Adds the partial bit processing - note we'll need to specify this is possible for ACVP testing for this and SHA3.


ounsworth:
We should probably port the bc-test-data and wycheproof test harnesses over from the mldsa crate as part of this PR.

dghgit:
bc-test-data testing has been ported. Partial/Full-byte test vectors added to bc-test-data/crypto/sha2.

dghgit:
Now on #113

PR #89, opened 2026-08-26
Feature/sm3 impl

dghgit:
Chinese SM3 Digest, including HMAC. Implementation supports partial bytes, may need to ask around at OpenSSL about getting some test vectors though.


ounsworth:
Thanks David!
On a 2-minute scroll-through, this code looks great, style-wise.
But to review this PR properly, including checking that the tests cover all the edge-cases that they need to cover, I would need to first read the SM3 spec cover-to-cover. Is there a public English version available?

So I think this PR will need to sit until I or someone else has time to review it deeply. Or maybe we could merge it into an `experimental/` folder instead of the `crypto/` folder with a readme about the review and testing needed before moving it to `crypto/`, disconnect it from the CLI and Factory, etc.

Thoughts?

ounsworth:
wycheproof has test vectors for sm4, but not sm3, so I'm also not sure where to get a good set of vectors.

dghgit:
English version of the spec is here:

http://www.gmbz.org.cn/upload/2024-11-18/1731899426565012428.pdf

Sadly all the examples are octet aligned.

There's also this:

https://datatracker.ietf.org/doc/draft-sca-cfrg-sm3/01/

For some unknown reason all the GM related standards submitted to the IETF ran out of puff. Attempts to contact the authors resulted in failure. I think one of the examples in the draft might have a typo in it as well, but BC's original implementation was based on the IETF draft.

If it's any further help, the SM3 implementation is based on BC's. The Rust implementation has also been tested against OpenSSL's, although note neither of these implementations support partial bytes. Both BC's and OpenSSL's implementations of SM3 have had extensive use at this point. I think we'll need to talk to the OpenSSL team concerning test vectors for non-byte aligned data. The implementation seems to be coherent with the spec, section 5.2 of the document shows the padding mechanism that's used.

crypto/factory/tests/hash_factory_tests.rs
```diff
@@ -56,6 +56,26 @@ mod hash_factory_tests {
...
+                assert_eq!(h.output_len(), 32);
+                assert_eq!(h.block_bitlen(), 512);
+                assert_eq!(
+                    h.hash(&DUMMY_SEED[..512]),
+                    b"\xb2\x1f\x83\x0d\xca\x06\xbe\x8b\x67\x8c\xf9\x87\xf2\x6b\x9a\x43\x6e\x1b\x42\x79\x63\xb4\x45\x03\x32\xf0\x12\x70\xbd\x2d\xf7\x5c"
+                );
+                let h = HashFactory::new(name).unwrap();
+                assert_eq!(
+                    h.hash(b"abc"),
+                    b"\x66\xc7\xf0\xf4\x62\xee\xed\xd9\xd1\xf2\xd4\x6b\xdc\x10\xe4\xe2\x41\x67\xc4\x87\x5c\xf2\xf7\xa2\x29\x7d\xa0\x2b\x8f\x4b\xa8\xe0"
```
npajkovsky: OpenSSL have more test vectors if you need 

https://github.com/openssl/openssl/blob/master/test/recipes/30-test_evp_data/evpmd_sm3.txt

kriskwiatkowski:
FWIW: I've implemented it long time ago also, in Go, should you wish to test against some other implementation. The code has some test vectors too:
https://github.com/kriskwiatkowski/nobs/blob/master/hash/sm3/sm3_test.go

dghgit:
Now on #113

PR #90, opened 2026-08-26
sha256: add AArch64 SHA-2 crypto-extension hardware backend

npajkovsky:
Adds an opt-in (`asm` feature) hardware compression backend for SHA-256 on little-endian aarch64, using the ARMv8 Cryptographic Extension (SHA256H/SHA256H2/SHA256SU0/SHA256SU1) via core::arch intrinsics. Runtime support is detected with
is_aarch64_feature_detected!, falling back to the portable scalar implementation everywhere else.

- New crypto/sha2/src/asm/ module: one submodule per algorithm (asm::sha256), dispatching by (target_arch, target_endian) at its own `mod` declaration; try_compress is always callable and is a no-op on targets with no backend, so callers never need their own #[cfg].
- #![forbid(unsafe_code)] moves from the crate root into sha256.rs/ sha512.rs (unconditional, independent of the `asm` feature); the crate root drops to #![deny(unsafe_code)] and asm/mod.rs grants itself the sole #![allow(unsafe_code)] exception, since `forbid` can never be overridden by a nested #[allow].
- Forwards the feature through the workspace Cargo.toml as `sha2-asm`.
- Extends the sha256 benchmark with a 1KiB-chunked and a 1MiB one-shot case alongside the existing 16KiB one-shot.

Benchmark (16KiB one-shot, criterion):
  baseline (scalar):    33.147 µs  ( 471.38 MiB/s)
  asm (aarch64 SHA-2):   4.757 µs  (3.2077 GiB/s)   ~7x faster

Assisted-by: Claude:sonnet-5

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -53,3 +53,7 @@ bouncycastle-mlkem-lowmemory.workspace = true
 bouncycastle-rng.workspace = true
 bouncycastle-sha2.workspace = true
 bouncycastle-sha3.workspace = true
+
+[features]
+# Forwarded opt-in for bouncycastle-sha2's hardware SHA-256 acceleration.
+sha2-asm = ["bouncycastle-sha2/asm"]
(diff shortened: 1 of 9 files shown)


npajkovsky:
I'll park the work in my local branch. If anyone is interested, I'll resubmit the work.

PR #94, opened 2026-08-27
sha3, rng cleanup

ounsworth:
Break up of https://github.com/bcgit/bc-rust/pull/87, so this is actually dgh's submission. So I will be reviewing / approving.

Mechanical cleanups, no behaviour change.

 use core:: imports, copy Copy types, drop a redundant zeroize

* std::marker::PhantomData -> core:: in sha3.rs, and std::fmt -> core:: in hash_drbg80090a.rs. Neither needs std; both crates are working towards no_std.
* KeyType and SecurityStrength are Copy, so min()/max() results are dereferenced rather than .clone()d. This also lets the surrounding expressions be formatted as single calls.
* Hash::hash_out() and XOF::hash_xof_out() called output.fill(0) before delegating to a path that zeroizes the buffer itself (do_final_out() and squeeze_out() respectively, per the trait contract). Dropped the outer fill; hash_out() no longer needs `mut output` or a reborrow.

### DIFF
--- a/alpha_0.1.3_release_notes.md
+++ b/alpha_0.1.3_release_notes.md
@@ -3,3 +3,9 @@
 ## Major features
 
 ## Minor features / bug fixes
+
+Housekeeping:
+
+* `no_std` progress: `std::marker::PhantomData` and `std::fmt` replaced with their `core::` equivalents in the SHA-3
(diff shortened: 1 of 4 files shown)


dghgit (approved):

dghgit:
Now in #113

PR #95, opened 2026-08-27
sha3: add Memory Usage bench and Security Considerations docs

ounsworth:
Break up of https://github.com/bcgit/bc-rust/pull/87, so this is actually dgh's submission. So I will be reviewing / approving.

QUALITY_AND_STYLE.md requires every primitive crate's docs to carry a "Memory Usage" section with a stack-usage table and, usually, a "Security Considerations" section; SHA-3 had neither.

The Memory Usage table's figures come from a new
mem_usage_benches/bench_sha3_mem_usage.rs, following the existing bench_mldsa_mem_usage / bench_mlkem_mem_usage: print_struct_sizes() reports size_of for SHA3_224..SHA3_512 and SHAKE128/256 (all 440 bytes, since every variant shares the same Keccak-f[1600] sponge) and SUSPENDED_SHA3_STATE_LEN (415). The remaining entry points -- one-shot hash, streaming, XOF squeeze, suspend/resume -- exist for valgrind --tool=massif peak-stack measurement, which criterion cannot do.

Security Considerations cites FIPS 202 Appendix A.1 for the strengths and A.2 for the XOF prefix property (SHAKE128(m, 32) is a prefix of SHAKE128(m, 64)), and notes that the sponge state is held in Secret and zeroized on drop.

### DIFF
--- a/alpha_0.1.3_release_notes.md
+++ b/alpha_0.1.3_release_notes.md
@@ -3,3 +3,10 @@
 ## Major features
 
 ## Minor features / bug fixes
+
+SHA-3 documentation:
+
+* Crate docs gained "Memory Usage" and "Security Considerations" sections.
+* New `mem_usage_benches/bench_sha3_mem_usage.rs` reports `size_of` for the SHA-3 / SHAKE objects (440 bytes) and the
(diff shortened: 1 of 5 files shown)


dghgit (approved):

dghgit:
Now in #113

PR #97, opened 2026-08-30
Part 4: Add Padding trait and bouncycastle-padding crate (PKCS7, Padded{En,De}cryptor)

dghgit:
Note this also includes a mock cipher to test with. We should replace that with AES when the actual implementation arrives. Rebased on release/0.1.3alpha, someone please review!

**Claude Summary** follows:

Adds the padding layer that the block-aligned `BlockCipherEncryptor` / `BlockCipherDecryptor` traits deliberately leave out.

## `core`

- `Padding<const BLOCK_LEN>` trait: in-place `pad(block, data_len)` and constant-time `unpad(block) -> data_len`.
- `PaddingError { DataLengthTooLong, InvalidPadding }`, plus a `PaddingError` variant (with `From`) on `SymmetricCipherError`.

## `crypto/padding` (new crate `bouncycastle-padding`, `no_std`, no `unsafe`)

- **`PKCS7`** — RFC 5652 §6.3 padding for any block length `1..=255`, enforced at compile time via an inline `const` assert. `unpad` visits every byte with `Condition<i64>` mask arithmetic from `bouncycastle-utils` and has a single public decision point, so neither timing nor error detail leaks a padding oracle. Spec text was taken from the downloaded RFC, not recall.
- **`PaddedEncryptor<E, P>` / `PaddedDecryptor<D, P>`** — adapt any `BlockCipherEncryptor` / `BlockCipherDecryptor` to arbitrary-length data. Streaming `do_update_out` / `do_final(self)` with exact `update_out_len` helpers, and one-shot `encrypt_out` / `decrypt_out`. Whole blocks go to the inner cipher in groups via `as_chunks` (no unwraps, no copies); the buffered partial plaintext block lives in a `Secret`. The decryptor withholds one complete block until `do_final`, since only the last block carries padding.
- Tests: PKCS7 round-trips for every length at block sizes 1/8/16/255 checked literally against the RFC rule, the RFC's worked strings, and malformed-padding rejection at every byte position. Adapter tests drive the code with a toy XOR-CBC cipher (the workspace's first implementor of the new traits) — every length, ten chunkings in both directions, the one-block lag, tampering, malformed lengths, buffer sizing, wrong `KeyType` — and run it through `TestFrameworkBlockCipher`.
- Criterion bench; registered in the workspace and re-exported as `bouncycastle::padding`.

## `core-test-framework` fix

Running the framework against the toy cipher exposed a latent bug: the security-strength loops in the symmetric/block/AEAD suites `unwrap`ped `set_security_strength`, whose comment claimed `do_hazardous_operations` bypasses the key-length guard. It doesn't (`key_material.rs`), so any key under 32 bytes panicked the suite. The loops now skip strengths the key cannot carry and the comment is corrected.

## Not included

No CLI subcommand — there is no real block cipher to wire it to yet; it belongs with the first AES/CBC crate.

## Verification

`cargo test --workspace`, `cargo doc` (no broken links), `cargo fmt --check`, `cargo bench -p bouncycastle-padding --no-run` all clean. `quality_stats.sh ./crypto/padding`: 322 code / 122 doc / 347 test lines.

### DIFF
--- a/Cargo.toml
+++ b/Cargo.toml
@@ -22,6 +22,7 @@ bouncycastle-mlkem = { path = "./crypto/mlkem" }
 bouncycastle-mlkem-lowmemory = { path = "./crypto/mlkem-lowmemory" }
 bouncycastle-mldsa = { path = "./crypto/mldsa" }
 bouncycastle-mldsa-lowmemory = { path = "./crypto/mldsa-lowmemory" }
+bouncycastle-padding = { path = "./crypto/padding" }
 bouncycastle-rng = { path = "./crypto/rng" }
 bouncycastle-sha2 = { path = "./crypto/sha2" }
 bouncycastle-sha3 = { path = "./crypto/sha3" }
@@ -55,6 +56,7 @@ bouncycastle-mldsa-lowmemory.workspace = true
 bouncycastle-mlkem.workspace = true
 bouncycastle-mlkem-lowmemory.workspace = true
 bouncycastle-modes.workspace = true
+bouncycastle-padding.workspace = true
 bouncycastle-rng.workspace = true
 bouncycastle-sha2.workspace = true
 bouncycastle-sha3.workspace = true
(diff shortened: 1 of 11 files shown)


dghgit:
Now on #113

PR #99, opened 2026-08-31
AES Low-Memory

officialfrancismendoza:
Independent AES low-memory crate to handle (#98)

PR #105, opened 2026-09-02
Part 2: aes-lowmemory: constant-time, table-free bit-sliced AES

dghgit:
Part 2 of 3, stacked on Part 1. Content is identical to #99 as reviewed and merged (commits `75789c9`..`f119c2a`, by @officialfrancismendoza).

- New crate `crypto/aes-lowmemory` (`bouncycastle-aes-lowmemory`): AES-128/192/256 as a raw keyed permutation, bit-sliced over eight `u32` planes with the 113-gate Boyar-Peralta S-box circuit -- no lookup tables, no secret-dependent branches or indexing, including in the key schedule.
- One unmodified key schedule serves both directions (FIPS 197 Algorithm 3); 176 / 208 / 240 B of state, pinned by test and by `mem_usage_benches/bench_aes_mem_usage`.
- Verified against FIPS 197 Appendix A (every schedule word) and B, an exhaustive S-box / inverse S-box check against Tables 4 and 6, SP 800-38A F.1, and 2138 NIST ACVP `ACVP-AES-ECB` cases from bc-test-data.
- Workspace, umbrella re-export, criterion benches, release note.

Deliberately ships no CLI, factory entry or cipher-trait impls: a bare permutation can only offer ECB. Those arrive with Part 3.


ounsworth:
I am starting a review of this PR as part of the 6-PR stack.
David has asked that I review the entire stack, and then circle back to the beginning and clean up my comments before anyone else start reacting to them. So I'll comment again on this PR when I'm ready for my comments to be taken seriously. (I may keep my review in Draft mode so that only I can see it until I'm ready to release my comments for all PRs at the same time).

Note: a 3,500 line PR will take me a number of days to read and form a mental picture of.

crypto/aes-lowmemory/src/lib.rs
```diff
@@ -0,0 +1,175 @@
+//! A constant-time, table-free AES block cipher engine (NIST FIPS 197).
```
ounsworth: "engine" is a word that I know is used in bc-java, but I have not been using in bc-rust because "engine" implies that you have a "car" that can accept different engines. That is definitely true under the Java JCA architecture, but I don't think it's true here.

I have just been calling them "implementations".

Although, in this case, this might be an "implementation" that is abstracted away by a mode (CBC, GCM, etc)?

I guess the question will be whether this crate is the public interface that we expect users to import, or whether it is a utility crate supporting public exports in a different crate. My intuition would be that this crate should contain the exports:

```rust
pub type AES_CBC_128 = ...;
pub type AES_GCM_128 = ...;
```
and it uses the Modes (and Padding) crates as internal utility crates, but I'll reserve judgment on that until I've reviewed the entire PR stack.

crypto/aes-lowmemory/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+#![allow(private_bounds)]
+
+mod aes;
+mod bitslice;
+mod round;
+mod sbox;
+mod schedule;
+
+pub use aes::{Aes, Aes128, Aes192, Aes256, BLOCK_LEN};
+pub use bitslice::Block;
```
ounsworth: I will hold off on actually changing this until I have cycled through the rest of the 6 PRs in this stack.

But on principle, my intuition is that the concept of a "block" is internal detail and a user should be able to use AES without ever needing to encounter the concept of a block, so this should not leak through the public API.

crypto/aes-lowmemory/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+// be added outside this crate; that is what triggers this lint.
+#![allow(private_bounds)]
+
+mod aes;
+mod bitslice;
+mod round;
+mod sbox;
+mod schedule;
+
+pub use aes::{Aes, Aes128, Aes192, Aes256, BLOCK_LEN};
```
ounsworth: I suggest that this crate is the right place to expose the pub types for AES_CBC_128, AES_GCM_256, etc, in fact I would argue that those should be the _only_ public APIs that this crate exposes because those are the only way that we want end users using AES.

crypto/aes-lowmemory/src/schedule.rs
```diff
@@ -0,0 +1,461 @@
...
+/// words are held little-endian here, so the word `Rcon[j]` is just this byte. Indexing is shifted
+/// by one against the spec: `RCON[j - 1]` is the spec's `Rcon[j]`, since the spec counts from 1.
+const RCON: [u32; 10] = [0x01, 0x02, 0x04, 0x08, 0x10, 0x20, 0x40, 0x80, 0x1b, 0x36];
+
+/// Prevents a fourth parameter set from being added outside this crate.
+///
+/// FIPS 197 Sec 6.1 defines exactly three: AES-128, AES-192 and AES-256. Because [`AesParams`]
+/// has this private supertrait, only the three types in this module can implement it, so no
+/// downstream crate can instantiate the cipher with an unapproved key length or round count.
+trait AesParamsSealed {}
```
ounsworth: The established pattern is to call this "AESParamsInternalTrait", see for example, [`MLKEMPrivateKeyInternalTrait`](https://github.com/bcgit/bc-rust/blob/main/crypto/mlkem/src/mlkem_keys.rs#L446), which is exactly the same pattern.

I know that "Sealed" is used as the example trait in all the Rust Lang docs for this "private bounds" pattern, but please make this PR consistent with the existing codebase by renaming this, and if you feel strongly then you can open a global refactoring ticket about "*Sealed" vs "*Internal".

crypto/aes-lowmemory/src/schedule.rs
```diff
@@ -0,0 +1,461 @@
...
+
+/// AES-128 parameters: 16-byte key, `Nk` = 4, `Nr` = 10 (FIPS 197 Sec 6.1).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Aes128Params;
+/// AES-192 parameters: 24-byte key, `Nk` = 6, `Nr` = 12 (FIPS 197 Sec 6.1).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Aes192Params;
+/// AES-256 parameters: 32-byte key, `Nk` = 8, `Nr` = 14 (FIPS 197 Sec 6.1).
+#[derive(Debug, Clone, Copy, PartialEq, Eq)]
+pub struct Aes256Params;
```
ounsworth: schedule.rs is an extremely weird place to define these global constants for AES.
Any objection if I move them to aes.rs or lib.rs?

(for example, the equivalents are in lib.rs for sha2, but in mlkem.rs / mldsa.rs, so either of those two choices is ok.)

crypto/aes-lowmemory/src/aes.rs
```diff
@@ -0,0 +1,276 @@
...
+    /// The key must be tagged [`KeyType::SymmetricCipherKey`], must be exactly `P::KEY_LEN` bytes
+    /// of the buffer, and must carry a [`SecurityStrength`] at least equal to its own length --
+    /// which is what a key of this length from a correctly-instantiated RNG or KDF will have.
+    /// The checks exist to catch a key that arrived from somewhere it should not have: a seed
+    /// reused as a cipher key, or a 32-byte buffer holding material only derived at the 128-bit
+    /// strength.
+    ///
+    /// Takes `&dyn KeyMaterialTrait` so the three constructors, whose `KeyMaterial<N>` capacities
+    /// differ, can share one implementation.
+    fn validate(key: &dyn KeyMaterialTrait) -> Result<(), SymmetricCipherError> {
```
ounsworth, line 46: I think the use of `dyn` is unnecessary here, and adds complexity to the API.

dyn is for cases where you need to accept any struct that impls that trait, but in fact here the _only_ type of KeyMaterial that we will accept is a `KeyMaterial<P::KEY_LEN>`, so can we simplify this to 

```rust
fn validate(key: &KeyMaterial<P::KEY_LEN>)
```
?

That also reduces the fallability of this function since we can delete the key length check below since this is now checked at compile time.

I don't think this point is really a "design decision" since using `dyn` here adds extra complexity and extra runtime error handling and I don't think we gain anything from it.

crypto/aes-lowmemory/src/sbox.rs
```diff
@@ -0,0 +1,381 @@
...
+    let u0 = q[7];
+    let u1 = q[6];
+    let u2 = q[5];
+    let u3 = q[4];
+    let u4 = q[3];
+    let u5 = q[2];
+    let u6 = q[1];
+    let u7 = q[0];
+
+    // Top linear transformation (23 gates): the input basis change.
```
ounsworth: Personally, I don't feel comfortable with these annotations because they are making security claims that are uncited and I don't know how to validate.

I'm going to take them out.

If we can source a citation for these, then I'll be happy to put them back in.

crypto/aes-lowmemory/src/lib.rs
```diff
@@ -0,0 +1,175 @@
+//! A constant-time, table-free AES block cipher engine (NIST FIPS 197).
+//!
+//! This crate provides the raw AES keyed permutation -- [`Aes128`], [`Aes192`] and [`Aes256`] --
+//! implemented as a Boolean circuit over bit-planes rather than as byte substitutions through a
+//! lookup table. That makes it both smaller and constant-time; see [Design](#design).
```
ounsworth, line 5: I'll leave this for now until I've done a full pass over the full stack of PRs, but I feel that exposing APIs for something called "AES" that is actually dangerous to use, is handing users rope to hang themselves with.

My feeling is that this crate, the `bouncycastle-aes` namespace is where people will go to look for usable AES primitives, not in a `bouncycastle-modes` namespace, cause that's not an intuitive place to look for AES. So I feel like this crate should use the `bouncycastle-modes`, `bouncycastle-padding` etc crates, and this crate should expose the `AES_CBC_128, AES_GCM_256, etc` primitives and *not* the raw unsafe `AES` ECB primitive (or, maybe it should, but type aliased to `AES_ECB`, and then stick the appropriate warnings on that.

crypto/aes-lowmemory/src/schedule.rs
```diff
@@ -0,0 +1,461 @@
...
+    const KEY_LEN: usize;
+    /// `Nk`, the key length in 32-bit words: 4, 6 or 8 (FIPS 197 Sec 6.1).
+    const NK: usize;
+    /// `Nr`, the number of rounds: 10, 12 or 14 (FIPS 197 Sec 6.1).
+    const NR: usize;
+    /// The algorithm name, as reported by `Algorithm::ALG_NAME`.
+    const ALG_NAME: &'static str;
+    /// `[u32; 4 * (NR + 1)]` -- the compressed schedule. See the module docs.
+    type Schedule: ZeroizablePrimitive + AsRef<[u32]> + AsMut<[u32]>;
+}
```
ounsworth, line 42: Why are these defined in schedule.rs?
Shouldn't this be in aes.rs?

Also, I really like this pattern, much cleaner than what I have in mldsa / mlkem.
NOTE-TO-SELF: go refactor the mess of params in MLDSA / MLKEM onto this pattern.

crypto/aes-lowmemory/src/aes.rs
```diff
@@ -0,0 +1,276 @@
...
+/// AES-192: 24-byte key, 12 rounds (FIPS 197 Sec 6.1).
+pub type Aes192 = Aes<Aes192Params>;
+/// AES-256: 32-byte key, 14 rounds (FIPS 197 Sec 6.1).
+pub type Aes256 = Aes<Aes256Params>;
+
+impl<P: AesParams> Aes<P> {
+    /// Checks a key is fit to use before it is expanded.
+    ///
+    /// The key must be tagged [`KeyType::SymmetricCipherKey`], must be exactly `P::KEY_LEN` bytes
+    /// of the buffer, and must carry a [`SecurityStrength`] at least equal to its own length --
```
ounsworth: I think SecurityStrength equal to its own length is the wrong requirement here -- I think it should be "equal to the security strength of the AES algorithm being instantiated", which we have available in `P::MAX_SECURITY_STRENGTH` ... although in practice those probably end up being the same, but I would phrase the requirement in terms of the algorithm, not in terms of the key, seems less brittle.

crypto/aes-lowmemory/src/lib.rs
ounsworth: This is sortof the same question as above. I should combine them before presenting to David / Francis.

crypto/aes-lowmemory/src/aes.rs
```diff
@@ -0,0 +1,276 @@
...
+        debug_assert_eq!(*block, discard, "the two interleaved halves must agree");
+    }
+
+    /// Decrypts one block in place. See [`Aes::encrypt_block`] for the two-blocks-at-once caveat.
+    pub fn decrypt_block(&self, block: &mut Block) {
+        let mut q = pack(block, block);
+        self.decrypt2(&mut q);
+        let mut discard = [0u8; BLOCK_LEN];
+        unpack(&q, block, &mut discard);
+        debug_assert_eq!(*block, discard, "the two interleaved halves must agree");
```
ounsworth, line 138: That debug_assert is clever!

crypto/aes-lowmemory/src/aes.rs
```diff
@@ -0,0 +1,276 @@
...
+    ///
+    /// # Errors
+    /// * [`KeyMaterialError::InvalidKeyType`] if the key is not [`KeyType::SymmetricCipherKey`].
+    /// * [`KeyMaterialError::InvalidLength`] if the key is not 16 bytes long.
+    /// * [`KeyMaterialError::SecurityStrength`] if the key carries a strength below 128 bits.
+    pub fn new(key: &KeyMaterial<16>) -> Result<Self, SymmetricCipherError> {
+        Self::validate(key)?;
+        Ok(Self { schedule: expand::<Aes128Params>(key.ref_to_bytes()) })
+    }
+}
```
ounsworth, line 161: I suspect that these three length-specific constructors can be combined into a single generified version:

```rust
impl<P: AESParams> AES<P> {
    /// Expands a key into an AES-128 schedule.
    ///
    /// # Errors
    /// * [`KeyMaterialError::InvalidKeyType`] if the key is not [`KeyType::SymmetricCipherKey`].
    /// * [`KeyMaterialError::SecurityStrength`] if the key carries a strength below 128 bits.
    pub fn new(key: &KeyMaterial<P::KEY_LEN>) -> Result<Self, SymmetricCipherError> {
        Self::validate(key)?;
        Ok(Self { schedule: expand::<P>(key.ref_to_bytes()) })
    }    
}
```

ounsworth: Also, the current docstring states:

```
    /// * [`KeyMaterialError::InvalidLength`] if the key is not 16 bytes long.
```

but I actually don't think this error state is triggerable because you're requiring a `&KeyMaterial<16>` which is already required to be 16 bytes by the compiler. Fix: just remove this line from the docstring.

I'm surprised that mutants didn't catch that the length check in `validate` is unreachable code. 🤔 That actually could use some digging.

crypto/aes-lowmemory/src/aes.rs
```diff
@@ -0,0 +1,276 @@
...
+    const ALG_NAME: &'static str = Aes256Params::ALG_NAME;
+    const MAX_SECURITY_STRENGTH: SecurityStrength = SecurityStrength::_256bit;
+}
+
+impl<P: AesParams> core::fmt::Debug for Aes<P> {
+    /// Prints the algorithm name only. The key schedule is secret and is never formatted.
+    fn fmt(&self, f: &mut core::fmt::Formatter<'_>) -> core::fmt::Result {
+        f.write_str(P::ALG_NAME)
+    }
+}
```
ounsworth, line 199: If we're doing `core::fmt::Debug`, should we also do `core::fmt::Display`? Everywhere else in the library, I implement them as a pair.

crypto/aes-lowmemory/src/lib.rs
```diff
@@ -0,0 +1,175 @@
...
+//! assert_eq!(block, [0x39, 0x25, 0x84, 0x1d, 0x02, 0xdc, 0x09, 0xfb,
+//!                    0xdc, 0x11, 0x85, 0x97, 0x19, 0x6a, 0x0b, 0x32]);
+//!
+//! // The same value decrypts, from the same schedule -- there is no separate decryptor.
+//! aes.decrypt_block(&mut block);
+//! assert_eq!(block, [0x32, 0x43, 0xf6, 0xa8, 0x88, 0x5a, 0x30, 0x8d,
+//!                    0x31, 0x31, 0x98, 0xa2, 0xe0, 0x37, 0x07, 0x34]);
+//! ```
+//!
+//! ## Two blocks at a time
```
ounsworth, line 45: Note-to-self, cause I'll probably figure this out as I look at the rest of the PRs in this stack, but what is the use for this? I would assume that the things that get used in the wild (AES_CBC, AES_GCM) can't use this 2-block mode?

crypto/aes-lowmemory/src/bitslice.rs
```diff
@@ -0,0 +1,210 @@
...
+    /// `cl` and `ch` are complementary, and `s` is exactly the field width, so in each returned
+    /// word the two combined operands occupy disjoint bits: `(x & cl)` and `(y & cl) << s` cannot
+    /// both be set in the same position. `|` and `^` therefore compute the same function here,
+    /// which is why `cargo mutants` reports the `| -> ^` mutants in this function as surviving --
+    /// they are equivalent programs. `test_ortho_is_an_involution` and
+    /// `test_layout_matches_the_documented_table` are what actually pin this code.
+    #[inline(always)]
+    fn swap(cl: u32, ch: u32, s: u32, x: u32, y: u32) -> (u32, u32) {
+        ((x & cl) | ((y & cl) << s), ((x & ch) >> s) | (y & ch))
+    }
```
ounsworth, line 36: I find this definition, even with the block comment, to be hard to understand. Can we find more descriptive variable names for cl, ch, and s?

Can we add some sort of clear sentence of the form:

> Returns (x, y) when ___ and (y, x) when ____

?

The comment says "cl and ch are complimentary", but I think this is actually a pre-condition for the function to be correct? IE the caller must ensure that they are complimentary?

crypto/aes-lowmemory/src/schedule.rs
```diff
@@ -0,0 +1,461 @@
...
+    ortho(&mut q);
+    sbox(&mut q);
+    ortho(&mut q);
+    q[0]
+}
+
+/// KEYEXPANSION() (FIPS 197 Sec 5.2, Algorithm 2), returning the compressed bit-sliced schedule.
+///
+/// `key` must be exactly `P::KEY_LEN` bytes; [`crate::aes`] checks that before calling, so this
+/// cannot fail and takes no `Result`.
```
ounsworth, line 122: Then why not put that in the method signature as `key: &[8; P::KEY_LEN]` ?

crypto/aes-lowmemory/src/schedule.rs
```diff
@@ -0,0 +1,461 @@
+//! KEYEXPANSION() (FIPS 197 Sec 5.2, Algorithm 2) and the per-key-length parameters.
+//!
+//! # Storage
+//!
+//! The schedule is `4 * (Nr + 1)` words -- 44, 52 or 60 -- exactly as FIPS 197 Sec 5.2 defines
+//! it, so 176, 208 or 240 bytes. It is stored in a **compressed** bit-sliced form: because
+//! bit-slicing is a permutation of bits it does not change the size, and because both interleaved
+//! blocks are encrypted under the same key the two halves of a bit-sliced round key are
+//! identical, so only one of every pair of words needs keeping. [`round_key`] re-doubles a single
+//! round key onto the stack when the round loop needs it.
```
ounsworth, line 10: I don't understand this paragraph. It is telling me that this is compressed, but then in the same sentence telling me that bit-slicing does not change the size.

My understanding is that 44, 52, or 60 words is the usual size of the AES key schedule. So in what way is this compressed? Maybe this paragraph just needs to be re-phrased?

ounsworth (changes requested):
Main question: this is `aes_lowmemory` -- are we planning to also do a separate non-low-memory implementation, or should we just rename this to `aes` and be done with it?

(FYI: The branch that Francis and I were working on is also based on the bitsliced SLP_AES_113.txt SBox, so it would also be considered a low-memory implementation)


If the big-and-slow implementation is the EqInvCipher() from §5.3.5, then I think we could easily add that as an additional API in this crate -- it's not like mldsa vs mldsa_lowmemory where basically every function in the crate needed to change. Although maybe it would be cleaner from an API perspective to present them as separate implementations? If the `aes` and `aes_lowmemory` share the same forward Cipher() direction, then we could have one crate depend on the other so at least we're not duplicating that code.

ounsworth:
While reviewing this PR, I prepared commit 736b0ac4 which mainly take a large pair of scissors to the docs so that the next person to look at this doesn't have to spend as much time reading.
It also makes some code changes, like renaming variables to match library convention, so upstream PRs in this stack will need to be rebased as we go.

dghgit:
Now on #113

PR #106, opened 2026-09-02
Part 3: AES CBC mode: BlockPermutation trait, bouncycastle-modes crate, CLI

dghgit:
Part 3 of 3, stacked on Part 2. Content is identical to #101 as reviewed and merged (commits `ff9fa67`..`70706e7`, by @officialfrancismendoza).

- `core`: new `BlockPermutation<KEY_LEN, BLOCK_LEN>` trait (`CIPH_K` / `CIPH^-1_K` of SP 800-38A Sec 5.1) with provided `encrypt_blocks2` / `decrypt_blocks2`; implemented by `Aes128/192/256`.
- New crate `crypto/modes` (`bouncycastle-modes`): `Cbc<P, Dir, KEY_LEN, BLOCK_LEN>` over any `BlockPermutation`, direction encoded in the type, IV generated (never accepted) per SP 800-38A Sec 5.3, pair-path decryption.
- `core-test-framework`: `TestFrameworkBlockPermutation`; fix to the `TestFrameworkBlockCipher` security-strength loop.
- CLI: `aes128-cbc` / `aes192-cbc` / `aes256-cbc` `encrypt|decrypt`, IV carried as the first block of the stream; 16 subprocess tests.
- Verified against all six SP 800-38A F.2 vector sets and 2150 NIST ACVP `ACVP-AES-CBC` cases; Appendix D error propagation tested exhaustively.

Follow-ups already prepared on `feature/aes-tweaks` (to be opened once this lands): `AES_CBC_128/192/256` aliases, flat-array one-shots and streaming methods, and a simpler CLI reader.

PR #107, opened 2026-09-02
Part 1: core: split BlockCipher into block-aligned BlockCipherEncryptor/Decryptor

dghgit:
Part 1 of 3 of the block-cipher stack, re-opened after `release/0.1.3alpha` was rolled back to `c026339` so the three pieces can land as separate, linear PRs. Content is identical to #96 as previously reviewed and merged (commits `e537e23`..`b770f56`).

- Split `BlockCipher` into `BlockCipherEncryptor` / `BlockCipherDecryptor` (mirroring `KEMEncapsulator` / `KEMDecapsulator`), with a minimal `BlockCipher` supertrait carrying `MAX_SECURITY_STRENGTH`.
- Replace single-block `do_*_block[_out]` with multi-block `do_*_blocks[_out]<const N>` over `[[u8; BLOCK_LEN]; N]`; add `do_encrypt_init_rng`; remove `do_*_final` (padding is a separate layer).
- Provided one-shot methods; core-test-framework block cipher suite updated (N = 1 and N = 2).
- Release note included.

Merge order: this PR, then #105 (Part 2, AES low-memory), then #106 (Part 3, CBC mode), then #97 (padding). Note for whoever merges: land it by pushing to `origin` -- GitHub is a mirror and a GitHub-side merge gets reverted by the sync bot.

### DIFF
--- a/alpha_0.1.3_release_notes.md
+++ b/alpha_0.1.3_release_notes.md
@@ -7,3 +7,26 @@
 * bug fixes to the way SHA3/SHAKE handled absorbing and squeezing a partial final byte.
 * Design discussions about whether core::traits::XOF (in the abstract) should allow interleaving absorb -> squeeze ->
   absorb (ie "absorb-after-squeeze). Outcome: absorb-after-squeeze forbidden. Could be changed in the future.
+
+Block cipher traits (PR #96):
+
+* The single `BlockCipher` streaming trait is split into `BlockCipherEncryptor` and `BlockCipherDecryptor` (mirroring
+  `KEMEncapsulator` / `KEMDecapsulator`) so the direction is encoded in the implementing type. A minimal `BlockCipher`
+  supertrait carries the shared `MAX_SECURITY_STRENGTH`; the `SymmetricCipher` one-shot API is no longer a supertrait.
+* The single-block `do_{en,de}crypt_block[_out]` methods are replaced by multi-block
+  `do_{en,de}crypt_blocks[_out]<const N>`, taking `&[[u8; BLOCK_LEN]; N]` so the block count is compile-time and
+  input/output lengths cannot disagree.
+* `do_encrypt_init_rng(key, &mut dyn RNG)` is added alongside `do_encrypt_init`, matching the `encaps` / `encaps_rng`
+  pattern.
+* The `do_{en,de}crypt_final[_out]` methods are removed: the traits are now strictly block-aligned, and padding of
+  arbitrary-length data belongs to a separate `PaddedEncryptor` / `PaddedDecryptor` layer built on top.
+* One-shot static APIs are provided (default) methods implemented once in the traits -- `encrypt_blocks`,
+  `encrypt_blocks_rng`, `encrypt_blocks_out`, `encrypt_blocks_out_rng` on `BlockCipherEncryptor` and `decrypt_blocks`,
+  `decrypt_blocks_out` on `BlockCipherDecryptor` -- so every block-aligned mode gets the house-standard
+  take-data-return-result API at no cost to implementors.
+
+Testing:
+
+* The core-test-framework block cipher test now takes separate encryptor/decryptor type parameters, exercises N = 1 and
+  N = 2 (including mixed single/multi-block encrypt vs decrypt sequences), and checks the one-shots agree with the
+  streaming API and round-trip.
--- a/crypto/core/src/traits.rs
+++ b/crypto/core/src/traits.rs
@@ -130,6 +130,44 @@ pub trait BlockCipherEncryptor<
         plaintext: &[[u8; BLOCK_LEN]; N],
         ciphertext: &mut [[u8; BLOCK_LEN]; N],
     ) -> Result<usize, SymmetricCipherError>;
+
+    /// One-shot: encrypts `N` blocks under a fresh init. Returns the generated init data and the ciphertext.
+    fn encrypt_blocks<const N: usize>(
+        key: &KeyMaterial<KEY_LEN>,
+        plaintext: &[[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], [[u8; BLOCK_LEN]; N]), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init(key)?;
+        Ok((init_data, enc.do_encrypt_blocks(plaintext)?))
+    }
+    /// As [`BlockCipherEncryptor::encrypt_blocks`], but sources randomness from the provided RNG.
+    fn encrypt_blocks_rng<const N: usize>(
+        key: &KeyMaterial<KEY_LEN>,
+        rng: &mut dyn RNG,
+        plaintext: &[[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], [[u8; BLOCK_LEN]; N]), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init_rng(key, rng)?;
+        Ok((init_data, enc.do_encrypt_blocks(plaintext)?))
+    }
+    /// One-shot: encrypts `N` blocks under a fresh init into the provided buffer.
+    /// Returns the generated init data and `N * BLOCK_LEN`.
+    fn encrypt_blocks_out<const N: usize>(
+        key: &KeyMaterial<KEY_LEN>,
+        plaintext: &[[u8; BLOCK_LEN]; N],
+        ciphertext: &mut [[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], usize), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init(key)?;
+        Ok((init_data, enc.do_encrypt_blocks_out(plaintext, ciphertext)?))
+    }
+    /// As [`BlockCipherEncryptor::encrypt_blocks_out`], but sources randomness from the provided RNG.
+    fn encrypt_blocks_out_rng<const N: usize>(
+        key: &KeyMaterial<KEY_LEN>,
+        rng: &mut dyn RNG,
+        plaintext: &[[u8; BLOCK_LEN]; N],
+        ciphertext: &mut [[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], usize), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init_rng(key, rng)?;
+        Ok((init_data, enc.do_encrypt_blocks_out(plaintext, ciphertext)?))
+    }
 }
 
 /// The decryption half of a block cipher's streaming API; see [`BlockCipherEncryptor`].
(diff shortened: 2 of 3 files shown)


alpha_0.1.3_release_notes.md
```diff
@@ -7,3 +7,26 @@
...
+* One-shot static APIs are provided (default) methods implemented once in the traits -- `encrypt_blocks`,
+  `encrypt_blocks_rng`, `encrypt_blocks_out`, `encrypt_blocks_out_rng` on `BlockCipherEncryptor` and `decrypt_blocks`,
+  `decrypt_blocks_out` on `BlockCipherDecryptor` -- so every block-aligned mode gets the house-standard
+  take-data-return-result API at no cost to implementors.
+
+Testing:
+
+* The core-test-framework block cipher test now takes separate encryptor/decryptor type parameters, exercises N = 1 and
+  N = 2 (including mixed single/multi-block encrypt vs decrypt sequences), and checks the one-shots agree with the
+  streaming API and round-trip.
```
ounsworth, line 32: Cut this down to a single line.
Or maybe just remove the release note for this PR entirely because the entire concept of symmetric ciphers is new in 0.1.3, so the release note will end up being:

* Implemented the following block ciphers, and the core traits to go along with them.
  * AES
  * AES_lowmemory
  * ASCON
  * ...

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
     ) -> Result<usize, SymmetricCipherError>;
 }
 
-/// The basic functions of a block cipher.
+/// Metadata shared by [`BlockCipherEncryptor`] and [`BlockCipherDecryptor`].
+pub trait BlockCipher {
+    /// Maximum security strength supported by the algorithm; keys tagged with a lower strength are
+    /// rejected by the `_init` constructors.
+    const MAX_SECURITY_STRENGTH: SecurityStrength;
+}
```
ounsworth, line 88: This is a literal duplicate of Algorithm. Why add a new trait instead of bounding BlockCipherEncryptor on Algorithm instead of on BlockCipher?

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
-    /// Encrypts the final block of plaintext.
-    fn do_encrypt_final(
+    /// As [`BlockCipherEncryptor::do_encrypt_init`], but sources randomness from the provided RNG.
+    fn do_encrypt_init_rng(
+        key: &KeyMaterial<KEY_LEN>,
+        rng: &mut dyn RNG,
+    ) -> Result<(Self, [u8; INIT_DATA_LEN]), SymmetricCipherError>;
+    /// Encrypts `N` consecutive blocks of plaintext. A sequence of calls is equivalent to one call over
+    /// the concatenation.
+    fn do_encrypt_blocks<const N: usize>(
```
ounsworth, line 123: This seems to me like an odd way to define this API.

How often is your application going to know N (ie the size of the plaintext) at compile-time?

I also still don't love forcing the user to break their plaintext into an array-of-arrays to match the block size -- something about that feels like leaking implementation detail through the abstraction layer. This should just take a plaintext as a simple &[u8].

I suggest that the better way to define this API would be:

```rust
    /// Encrypts multiple consecutive blocks of plaintext. A sequence of calls is equivalent to one call over
    /// the concatenation.
    /// The provided `plaintext` must be an even number of blocks; ie a multiple of [`Self::BLOCK_LEN`],
    /// or else a [`SymmetricCipherError::LengthError`] is returned.
    /// Since block ciphers guarantee that ciphertext and plaintext are the same size, the ciphertext
    /// is written to the provided `plaintext` buffer.
    /// Returns the number of bytes written to `plaintext`.
    fn do_encrypt_blocks(&mut self, plaintext: &mut [u8]) -> Result<usize, SymmetricCipherError>;
```

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
         &mut self,
-        plaintext: &[u8; BLOCK_LEN],
-    ) -> Result<[u8; BLOCK_LEN], SymmetricCipherError>;
-    /// Encrypts the final block of plaintext and writes the ciphertext to the provided buffer.
-    fn do_encrypt_final_out(
+        plaintext: &[[u8; BLOCK_LEN]; N],
+    ) -> Result<[[u8; BLOCK_LEN]; N], SymmetricCipherError>;
+    /// Encrypts `N` consecutive blocks of plaintext into the provided buffer. Returns `N * BLOCK_LEN`.
+    fn do_encrypt_blocks_out<const N: usize>(
```
ounsworth, line 128: If you take my suggestion to handle plaintext as a input/output argument, then this version can be deleted.

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
         &mut self,
-        plaintext: &[u8; BLOCK_LEN],
-        ciphertext: &mut [u8; BLOCK_LEN],
+        plaintext: &[[u8; BLOCK_LEN]; N],
+        ciphertext: &mut [[u8; BLOCK_LEN]; N],
     ) -> Result<usize, SymmetricCipherError>;
-    /// Constructor that begins a flow of the streaming API for decryption one block at a time.
+
```
ounsworth, line 133: Is there any case for a block cipher to have a `do_final()`?
I guess if none of the block ciphers that you're cooking need it, then that's that!

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
+    /// One-shot: encrypts `N` blocks under a fresh init. Returns the generated init data and the ciphertext.
+    fn encrypt_blocks<const N: usize>(
+        key: &KeyMaterial<KEY_LEN>,
+        plaintext: &[[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], [[u8; BLOCK_LEN]; N]), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init(key)?;
+        Ok((init_data, enc.do_encrypt_blocks(plaintext)?))
+    }
```
ounsworth, line 141: I suggest instead:

```rust
    /// One-shot: encrypts multiple blocks under a fresh init. Returns the generated init data and the ciphertext.
    ///
    /// The provided `plaintext` must be an even number of blocks; ie a multiple of [`Self::BLOCK_LEN`],
    /// or else a [`SymmetricCipherError::LengthError`] is returned.
    /// Since block ciphers guarantee that ciphertext and plaintext are the same size, the ciphertext
    /// is written to the provided `plaintext` buffer.
    /// Returns the number of bytes written to `plaintext`.
    fn encrypt(
        key: &KeyMaterial<KEY_LEN>,
        plaintext: &mut [u8],
    ) -> Result<([u8; INIT_DATA_LEN], usize), SymmetricCipherError> {
        let (mut enc, init_data) = Self::do_encrypt_init(key)?;
        Ok((init_data, enc.do_encrypt_blocks(plaintext)?))
    }
```

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
+    /// As [`BlockCipherEncryptor::encrypt_blocks`], but sources randomness from the provided RNG.
+    fn encrypt_blocks_rng<const N: usize>(
+        key: &KeyMaterial<KEY_LEN>,
+        rng: &mut dyn RNG,
+        plaintext: &[[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], [[u8; BLOCK_LEN]; N]), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init_rng(key, rng)?;
+        Ok((init_data, enc.do_encrypt_blocks(plaintext)?))
+    }
```
ounsworth, line 150: I suggest instead:

```rust
    /// Same as [`BlockCipherEncryptor::encrypt_blocks`], but sources randomness from the provided RNG.
    fn encrypt_blocks_rng<const N: usize>(
        key: &KeyMaterial<KEY_LEN>,
        rng: &mut dyn RNG,
        plaintext: &mut [u8],
    ) -> Result<([u8; INIT_DATA_LEN], usize), SymmetricCipherError> {
        let (mut enc, init_data) = Self::do_encrypt_init_rng(key, rng)?;
        Ok((init_data, enc.do_encrypt_blocks(plaintext)?))
    }
```

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
+    /// One-shot: encrypts `N` blocks under a fresh init into the provided buffer.
+    /// Returns the generated init data and `N * BLOCK_LEN`.
+    fn encrypt_blocks_out<const N: usize>(
```
ounsworth, line 153: Can delete if we make the above have an input/output argument.

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
+        key: &KeyMaterial<KEY_LEN>,
+        plaintext: &[[u8; BLOCK_LEN]; N],
+        ciphertext: &mut [[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], usize), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init(key)?;
+        Ok((init_data, enc.do_encrypt_blocks_out(plaintext, ciphertext)?))
+    }
+    /// As [`BlockCipherEncryptor::encrypt_blocks_out`], but sources randomness from the provided RNG.
+    fn encrypt_blocks_out_rng<const N: usize>(
```
ounsworth, line 162: Ditto

crypto/core/src/traits.rs
```diff
@@ -80,73 +80,138 @@ pub trait SymmetricCipher<const KEY_LEN: usize, const INIT_DATA_LEN: usize>: Alg
...
+        plaintext: &[[u8; BLOCK_LEN]; N],
+        ciphertext: &mut [[u8; BLOCK_LEN]; N],
+    ) -> Result<([u8; INIT_DATA_LEN], usize), SymmetricCipherError> {
+        let (mut enc, init_data) = Self::do_encrypt_init_rng(key, rng)?;
+        Ok((init_data, enc.do_encrypt_blocks_out(plaintext, ciphertext)?))
+    }
+}
+
+/// The decryption half of a block cipher's streaming API; see [`BlockCipherEncryptor`].
+pub trait BlockCipherDecryptor<
```
ounsworth, line 174: Ditto the changes to BlockEncryptor

ounsworth (changes requested):

crypto/core/src/traits.rs
ounsworth: Also, this is the first inline code in a trait. I honestly didn't know you could do that. I like it.

ounsworth:
Also, it's been bothering be for a while that the traits are not sorted alphabetically. This PR would be a good place to do that since it's just the symmetric cipher ones that are out-of-order.

crypto/core/src/traits.rs
dghgit: It was in SymmetricCipher. Algorithm would work though.

crypto/core/src/traits.rs
dghgit: No do_final is for tags and padding.

dghgit:
Now on #113

PR #109, opened 2026-09-02
Part 5: AES tweaks: AES_CBC_* aliases, flat-array one-shots and streaming, simpler CLI

dghgit:
Follow-ups to the block-cipher stack, stacked on #97 (padding). Three commits:

1. **`aes-lowmemory`: `AES_CBC_128` / `AES_CBC_192` / `AES_CBC_256` type aliases** over `bouncycastle_modes::Cbc`, generic in the direction marker, so callers never spell out the `KEY_LEN` / `BLOCK_LEN` const parameters (`AES_CBC_128::<Encrypting>::encrypt(&key, &msg)`). They live in the AES crate because the modes crate is deliberately cipher-agnostic; `aes-lowmemory` gains a dependency on `modes` (modes keeps its dev-dependency on `aes-lowmemory` -- a Cargo-permitted dev-dep cycle, verified by the modes test suite). Naming follows `HMAC_SHA256` / `HKDF_SHA256`. Each alias has a two-direction doctest, since Rust only checks an alias's bounds at a use site.

2. **`core`: the block-cipher one-shots take a flat `[u8; LEN]`.** `encrypt` / `encrypt_rng` / `encrypt_out` / `encrypt_out_rng` and `decrypt` / `decrypt_out` replace the block-shaped `encrypt_blocks*` / `decrypt_blocks*` from #96. `LEN % BLOCK_LEN == 0` is enforced **at compile time** by an inline `const` assertion at the instantiating call site -- no runtime length check and no error variant; a 47-byte array is a `compile_fail` doctest. Internally `LEN / BLOCK_LEN` is not nameable without `generic_const_exprs`, so the buffer is walked with `as_chunks`: pairs first (so a mode's two-block path is used), then the at-most-one remaining block.

3. **`core`: flat streaming methods replace the by-value block methods; simpler CLI.** `do_encrypt<LEN>` / `do_encrypt_out<LEN>` and `do_decrypt<LEN>` / `do_decrypt_out<LEN>` are provided; the by-value `do_*_blocks<N>` are removed. The one block-shaped method left is the implementor hook `do_*_blocks_out<N>` over `[[u8; BLOCK_LEN]; N]` -- it guarantees an implementation never sees a partial block and that in/out lengths agree at compile time, and serves runtime-count tails. An implementor now writes only `init[_rng]` and that hook. The CLI's `aes*-cbc` commands stream stdin into a flat 1 KiB buffer (`do_*_out::<1024>` for full chunks, `do_*::<16>` for the whole-block tail); the block staging buffer and partial-block carry logic are gone. Behaviour and error messages unchanged (16 CLI subprocess tests).

None of this has shipped (0.1.3 is unreleased), so the removals are not breaking changes; the #96 release-note bullet is updated to describe the final API. Verified: `cargo fmt --check` clean, full workspace green (92 suites / 586 tests), benches compile, zero rustdoc warnings.

Merge after #97. Land by pushing to `origin` (GitHub is a mirror).


dghgit:
Now on #113

Issue #103, opened 2026-09-01
AES LightEngine CFB Mode


officialfrancismendoza:
Continuing with the aes-lowmemory crate, implement CFB mode

PR #101, opened 2026-09-01
AES-lightengine CBC mode

officialfrancismendoza:
Continuing to build upon our low memory derivative of AES, CBC functionality and associated tests must be implemented. Added CBC mode, associated test vectors, and special CLI commands to test specific CBC variations. Continuing building upon the low memory derivative of AES, based on changes in (#98, #99) for AES lowmemory (#100).

Issue #100, opened 2026-08-31
AES LightEngine CBC Mode


officialfrancismendoza:
AES-CBC functionality based off of individual AES lowmemory crate (#98, #99)

officialfrancismendoza:
Merged to release/0.1.3alpha as of #101

Issue #98, opened 2026-08-31
AES Low-Memory


officialfrancismendoza:
A constant-time AES Low-Memory implementation. Also important preliminary step to add modes in a meaningful way, including GCM.

officialfrancismendoza:
Merged to release/0.1.3alpha as detailed in PR #99, so closing

PR #111, opened 2026-09-02
Part 6: Revamped AES Lightengine CFB mode from core API tweaks

officialfrancismendoza:
Building upon previous work on aes-lowmemory derivative crate (#98, #99) and CBC mode compatible with the aes-lowmemory derivative (#100, #101); specifically to implement **AES CFB mode** for the new aes-lowmemory crate and factoring in the new changes to the core API design.

This feature branch implements the lowmemory derivative of AES, with its corresponding CFB mode (#103), to be compatible with the new BlockCipherEncryptor / BlockCipherDecryptor split and BlockPermutation.


dghgit (approved):
Sanity check passed, ACVP passed.

Fabel 5.1 audit, passed.

officialfrancismendoza:
@dghgit checked post force-push, seems fine.

dghgit (approved):
There is an issue with this one, but we cannot address it just yet. CFB is actually a StreamCipher, so while the expression of the mode here is correct, the way it's having to be used isn't. We'll deal with that this week.

dghgit:
Now on  #113

PR #121, opened 2026-09-08
Refactor of HMACParams

ounsworth:
# Original issue:

Adds docs for cloning a hash function mid-stream so that you now have two hash objects that are identical up to this point, but you can feed different suffix input into before finalizing.

I put this next to the existing sections for `[Suspendable]`, since it's closely-related functionality. (and re-arranged order to put these before Memory Usage and Security Consideration sections).

# Extension

During development of that, I realized that the pub interface flip of HMAC and HKDF introduced in #122 meant that the way HMACParams was defined was now really clunky. A refactor of that lead to realizing that actually, SP 800-107r1 specifies the security strength of an HMAC higher than the collision strength of the underlying hash function, so I re-worked that as well.

### DIFF
--- a/alpha_0.1.3_release_notes.md
+++ b/alpha_0.1.3_release_notes.md
@@ -7,3 +7,7 @@
 * bug fixes to the way SHA3/SHAKE handled absorbing and squeezing a partial final byte.
 * Design discussions about whether core::traits::XOF (in the abstract) should allow interleaving absorb -> squeeze ->
   absorb (ie "absorb-after-squeeze). Outcome: absorb-after-squeeze forbidden. Could be changed in the future.
+* Re-arranged the HMAC and HKDF crates so that they are utility crates, and the pub types HMAC_SHA256, HMAC_SHA3_256,
+  HKDF_SHA256, and so on now live in the `bouncycastle_sha2::hmac`, `bouncycastle_sha3::hmac`, and
+  `bouncycastle_sha2::hkdf` namespaces. At the same time, adjustments were made to the MAX_SECURITY_STRENGTH of the HMAC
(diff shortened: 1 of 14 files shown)


dghgit:
Okay, so it looked fine when I looked at it earlier, but it looks like it's based on the old release/0.1.3alpha tip, so it's picked up the extra commits. It'll need to be re-based back to the current 0.1.3alpha tip. To be honest, I recommend re-basing it off #113 or #115, it will probably be easier and likely give a cleaner result.

Once the re-basing is done, this can be considered approved.

dghgit (changes requested):
It's been affected by the re-base. See comment at bottom of discussion page.

ounsworth:
@dghgit Right. We rebased release/0.1.3alpha to take the AES, SM3, and SHA512/t stuff off it. I have now rebased this to match.

While I was at it, I decided to have a go at simplifying how the HMACParams trait is structured, and that lead me to reading SP 800-107r1 §5.3.4 and I realized that it allows the HMAC's security strength to be higher
than the collision strength of the underlying hash function (because HMAC does not rely on collision resistance.)
I have re-done the MAX_SECURITY_STRENGTH of all the SHA2 and SHA3 instantiations under this new reading. I would like your close review of that part of this PR.
* Starting at crypto/sha2/src/hmac.rs:331
* Starting at crypto/sha3/src/hmac.rs:315

An artifact of #122 (merged earlier today), which moved the HMAC_SHA2 and HMAC_SHA3 pub types from the HMAC crate to the SHA2 / SHA3 crates is that the HMAC integration tests should have moved with them. That has been done on this PR as well.

dghgit:
Actually #122 caused quite a few merge conflicts, this would be better re-based off #118 (which will have merge conflicts, but at least you'll know the end result is what you are looking for).

Issue #128, opened 2026-09-10
Refactor ct::eq_bytes to use an unsafe read_votatile


ounsworth:
The current constant-time equality uses BlackBox, which is not a great solution.

the `bouncycastle-utils` crate is allowed to use Unsafe Rust. A better solution would be to do something with `read_volatile` and `write_volatile`.

PR #131, opened 2026-09-12
utils ct: replace black_box with a volatile optimization barrier in ct_eq_bytes and friends

ounsworth:
Survey of other libraries (from their current upstream sources): libsodium's `sodium_memcmp` uses a `volatile` accumulator and volatile input pointers, and subtle's `Choice::from` is an `#[inline(never)]` `read_volatile`, so both place the barrier inside the loop per byte. `constant_time_eq` and RustCrypto's `cmov` chunk into machine words and apply an inline-asm barrier (or `cmov`/`csel`) per word, which is the shape adopted here. Graviola writes the whole byte loop in inline asm on x86_64 and aarch64. OpenSSL's `CRYPTO_memcmp` reads through volatile pointers only, and BoringSSL's uses no barrier at all.

Assembly was inspected by Claude from release builds on x86_64, i686, thumbv7em (Cortex-M4), riscv32imac, wasm32, msp430 (16-bit) and avr-none (8-bit): in every case the loop is load, xor, or, one native-width store to a stack slot and one reload, with a byte-wise tail, no `bcmp`/`memcmp` call and no data-dependent branch, and the final `== 0` tests the volatile-loaded value.

Integration tests now sweep every length from 0 to 40 with a single bit flipped at every position, covering the word and tail paths on both sides of every boundary for 2, 4 and 8-byte words.

Co-Authored-By: Claude Fable 5

Closes #128

### DIFF
--- a/crypto/utils/src/ct.rs
+++ b/crypto/utils/src/ct.rs
@@ -560,52 +560,128 @@ where
     }
 }
 
-/// Rust doesn't guarantee that anything can truly be constant-time under all compilation targets
-/// and optimization levels. The following presents the standard constant-time shape.
+// ---------------------------------------------------------------------------------------------
+// Byte-slice comparison helpers
+//
+// The core idea here is that the data-dependent state is routed through
+// volatile memory accesses on every iteration. The documentation of
+// `core::ptr::read_volatile` / `write_volatile` states that they "are guaranteed to not be elided
+// or reordered by the compiler" and that a volatile read "will actually access memory and not
+// e.g. be lowered to reusing data from a previous read". So once the accumulator has been
+// volatile-written and volatile-read back, the optimiser holds no facts about its value, and in
+// particular cannot introduce either of the two early exits that would otherwise be legal:
+//
+//   * leaving the loop once the accumulator is non-zero, because the final `== 0` is already
+//     decided (only legal if the compiler can see that the zero test is the sole consumer), and
+//   * leaving the loop once the accumulator is all-ones, because further ORs cannot change it
+//     (legal regardless of the consumer, which is why the barrier must be *inside* the loop).
+// ---------------------------------------------------------------------------------------------
+
+/// As a performance optimization, we compare one machine-word at a time, which we assume to be the same
+/// as the size of a `usize`.
+type AccWord = usize;
+const ACC_BYTES: usize = size_of::<AccWord>();
+
+/// The cor operation that implements the optimization barrier by performing a
+/// `*acc |= diff`, performed as a volatile store followed by a volatile load, so that the
+/// compiler retains no knowledge of the accumulator's value afterwards. See the module comment
+/// above for why this is the barrier used in this file.
+#[inline(always)]
+fn volatile_or_assign(acc: &mut AccWord, diff: AccWord) {
+    // SOUNDNESS:
+    //   * We first write_volatile to the accumulator so that the compiler does not own the written value.
+    //   * We then read_volatile back into `*acc` so that the compiler does not own the value of `*acc`
+    //     on exit, meaning that any read on `*acc` after this exits cannot be elided either.
+    // SAFETY:
+    //  * `acc` is a `&mut AccWord`, so it is non-null, aligned for `AccWord`, and points at an
+    //    initialised `AccWord` inside a live Rust allocation (the caller's stack frame). It is
+    //    therefore valid for both reads and writes for the whole call, which is exactly the
+    //    precondition of `write_volatile` and `read_volatile`.
+    //  * The `&mut` is exclusive, so no other reference can observe or race with these accesses.
+    //  * `AccWord` is `Copy` with no drop glue, so overwriting it and bitwise-copying it back
+    //    out neither leaks nor double-drops anything.
+    //  * Every bit pattern is a valid `AccWord`, so the value read back is always initialised.
+    unsafe {
+        core::ptr::write_volatile(acc, *acc | diff);
+        *acc = core::ptr::read_volatile(acc);
+    }
+}
+
+/// Constant-time equality of two byte slices.
+///
+/// The runtime depends on the *lengths* of the inputs, which are treated as public, but not on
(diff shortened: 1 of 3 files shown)


dghgit:
Analysis courtesy of Opus 5. Some of these errors are extremely odd if this PR was generated using Fable - the LLMs do lend more weight to comments/text than code by default which could explain the mutant issue, but I'm not sure about the rest.

### 1. A killable mutant survives: `|` → `^` in `volatile_or_assign` (`crypto/utils/src/ct.rs:605`)

`cargo mutants` over the four changed functions gives 22 mutants, 20 caught, 2 missed. This one is **not** an OR/XOR equivalence of the kind we normally accept in crypto code — with `^` the accumulator *cancels*, so two differences that produce the same word-relative diff make unequal inputs compare equal. For example, flipping bit 0 of byte 0 and bit 0 of byte 8 gives `acc = 1 ^ 1 = 0`, and `ct_eq_bytes` returns `true`.

The new boundary sweep only ever flips a single bit at a time, so it cannot see this. Adding one case with two differences exactly one word apart (and, ideally, one pair within the byte-wise tail) closes the gap and kills the mutant.

### 2. The other missed mutant is genuinely equivalent

`|` → `^` at `crypto/utils/src/ct.rs:685` in `conditional_copy_bytes`. `mask` is all-ones or all-zeros, so exactly one of `a[i] & mask` and `b[i] & !mask` is non-zero and `|` is identical to `^`. Worth recording in the PR so nobody chases it later.

### 3. The release-notes bullet has three errors in two lines

```
* Refactored constant-time ct_equals_bytes () to use an optimization barrier based on unsafe read_volatile /
  write_volatile insead of core::mem::black_box.
```

- `ct_equals_bytes()` is not a function in this repo — the name is `ct_eq_bytes`.
- `insead` → `instead`.
- `core::mem::black_box` → `core::hint::black_box` (which is what the code actually replaced).

There is also a trailing space at the end of the second line.

### 4. `is_in_list` still carries the question this PR just answered

`crypto/utils/src/ct.rs:121` and `:262` (the `i64` and `i32` impls) still read:

> // Research question: is this actually constant-time?
> // A clever compiler might turn this into a short-circuiting loop.
> // A quick google search shows that rust doesn't have the ability to annotate specific code blocks
> // as no-optimize; the only option is to insert direct assembly.


That is precisely the saturating-accumulator early exit described in this PR's new banner comment, and `volatile_or_assign` is the answer to it. As merged, `ct.rs` would run two different constant-time disciplines side by side, with the older one still flagged as an open question and the last sentence ("the only option is to insert direct assembly") now plainly contradicted by this PR's own argument.

Either extend the barrier to those two loops — which needs a per-width analogue, since the accumulator there is `i64`/`i32` rather than `usize` — or, at minimum, replace the comment with a pointer to the new barrier plus a follow-up issue.

### 5. Minor wording

- `crypto/utils/src/ct.rs:582`: "The **cor** operation that implements the optimization barrier".
- Both new doc comments say "See the module comment above", but the block they refer to is a plain `//` banner rather than a `//!` module doc, so the reference does not resolve for anyone reading the rendered rustdoc.

### 6. The performance claim is unmeasured

"As a performance optimization, we compare one machine-word at a time" — `crypto/utils` has no `benches/` directory and none is added here, so the 8x reduction in barrier round-trips is asserted rather than shown. Not a blocker, but if the word chunking is being justified on performance grounds it is cheap to back it up with a criterion bench.

crypto/utils/src/ct.rs
```diff
@@ -560,52 +560,128 @@ where
...
+    //  * `u8` is `Copy` with no drop glue and every bit pattern is a valid `u8`, so the value
+    //    read back is initialised and nothing is leaked or double-dropped.
+    unsafe {
+        core::ptr::write_volatile(&mut mask, mask);
+        mask = core::ptr::read_volatile(&mask);
+    }
+
     for i in 0..LEN {
-        out[i] = core::hint::black_box(a[i] & mask) | core::hint::black_box(b[i] & !mask);
+        out[i] = (a[i] & mask) | (b[i] & !mask);
```
npajkovsky, line 685: This looks like `select()`.

crypto/utils/src/ct.rs
```diff
@@ -560,52 +560,128 @@ where
...
 
+    // SAFETY:
+    //  * `&mut mask` is a reference to an initialised, aligned `u8` local on this stack frame, so
+    //    it is valid for reads and writes for the duration of both calls.
+    //  * The reference is exclusive; nothing else can observe `mask` during the two accesses.
+    //  * `u8` is `Copy` with no drop glue and every bit pattern is a valid `u8`, so the value
+    //    read back is initialised and nothing is leaked or double-dropped.
+    unsafe {
+        core::ptr::write_volatile(&mut mask, mask);
+        mask = core::ptr::read_volatile(&mask);
```
npajkovsky, line 681: That seems too much for my taste.

OpenSSL defines `select()`

```c
static ossl_inline unsigned int constant_time_select(unsigned int mask,
    unsigned int a,
    unsigned int b)
{
    return (value_barrier(mask) & a) | (value_barrier(~mask) & b);
}
```

where the `value_barrier()` is defined:

```c
static ossl_inline unsigned int value_barrier(unsigned int a)
{
#if !defined(OPENSSL_NO_ASM) && defined(__GNUC__)
    unsigned int r;
    __asm__("" : "=r"(r) : "0"(a));
#else
    volatile unsigned int r = a;
#endif
    return r;
}
```

Its empty inline assembly uses tied input/output operands to hide the mask’s value from the optimizer.

If `asm` block is not an option, then `core::ptr::read_volatile`  should be enough.

I suggest implementing a `value_barrier` helper.

crypto/utils/src/ct.rs
ounsworth: It does look like the same core code as Condition<T>.select().

I guess you're suggesting to use Condition<>.select() here rather than duplicate?

npajkovsky: 💯

PR #120, opened 2026-09-08
Feat: Core AEAD Cipher Split

officialfrancismendoza:
Implements core AEAD cipher split (#119), based on #118 (`feature/xof-cshake`). Implements AEADCipherEncryptor and AEADCipherDecryptor, as well as cherry-picks commits from the ASCON PR (#21) so that ASCON is compatible — this PR supersedes #21 and implements #22.

PR #133, opened 2026-09-16
Simple ciphers: stream + symmetric cipher unification off the release branch, plus SHA-512/t for any t

dghgit:
Consolidates the stacked pair #113 (`feature/stream-cipher` → `release/0.1.3alpha`) and #115
(`feature/symmetric-cipher` → `feature/stream-cipher`) into a single branch off the release
branch, and adds the SHA-512/t work on top.

`feature/simple-ciphers` was built by branching from `release/0.1.3alpha` and merging
`feature/stream-cipher` then `feature/symmetric-cipher`; both merges fast-forwarded, so the
content below the two new commits is exactly what #113 and #115 already carried — this is a
re-packaging, not new review surface. #113 (+21,938) and #115 (+2,310) sum to roughly the
+23,499 here.

### New in this PR, on top of those two

**`core`: `SecurityStrength::from_bits` / `from_bytes` become `const fn`** (3912434) — two
keywords, no behaviour change. Needed so an associated const can derive a strength from a const
generic; split out ahead of the feature per CLAUDE.md's scope-of-changes rule.

**`sha2`: `SHA512t<T>` is usable for every `t` FIPS 180-4 s. 5.3.6 defines a hash for** (785dbef).
Previously only `T = 224` and `T = 256` had parameter impls, so nothing else could be named.

- **The t-value checks are back, and they are the spec's.** b11f8f6 had narrowed `sha512t_h0` to
  `100 <= t < 512` and emitted a fixed three digits, since only 224 and 256 were reachable. With
  arbitrary `t` reachable that would produce the `"0256"` spelling s. 5.3.6 explicitly forbids —
  and hence the wrong IV — for every `t` below 100. The one- and two-digit branches are restored
  and `check_t` asserts the section's own rule: positive, `< 512`, not 384.
- **One deviation, documented:** `t` must be a multiple of 8, because `Hash` is byte-oriented and
  a 100-bit digest has no representation here. BC Java's `SHA512tDigest` imposes the identical
  restriction, so the two libraries accept the same set of truncations.
- **Unapproved truncations are gated** the way `ElectronicCodeBook::ENCRYPTION_APPROVED` gates
  two-key TDEA: `SHA512tParams::<T>::FIPS_APPROVED` is public and `SHA512Internal::new` asserts it
  in an inline `const`, so an unapproved `t` is a compile error at the call site and
  `new_allow_unapproved_t()` is the deliberate way in. That also blocks `Default`, which keeps an
  unapproved truncation out of generic code by accident.
- `ALG_NAME`, `OUTPUT_LEN` and `MAX_SECURITY_STRENGTH` are derived from `t`, with `const`
  assertions pinning them to the values 224 and 256 previously had by hand.

### Verification

- `cargo test --workspace`: 943 passed, 0 failed (was 923). `cargo fmt --check` clean; no new
  clippy warnings. Both new commits build independently.
- Spec text read from a freshly downloaded FIPS 180-4, not from recall.
- **Test vectors cross-checked against BC Java's `SHA512tDigest`**, an independent implementation:
  eight truncations (8, 16, 24, 88, 96, 104, 264, 504) spanning all three decimal branches, over
  the FIPS 180-4 Appendix C messages plus the one-million-'a' case. The 224/256 rows in the same
  table match the NIST-published values.
- `cargo mutants` on the changed files: 259 mutants — 180 caught, 5 timeout-kills, 69 unviable,
  5 missed. All five missed are the pre-existing XOR/OR equivalences already annotated at their
  sites in `ch`, `maj` and `do_final_internal`; no new missed mutants.

### Note for reviewers

Adding `const` to a public `core` function is a forward compatibility commitment, and
`SHA512tParams` is currently its only caller. The alternative was a private copy of the rounding
ladder inside `sha2`, free to drift from the real one — happy to switch if the API-surface cost is
the greater worry.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


crypto/hmac/Cargo.toml
```diff
@@ -7,7 +7,7 @@ edition.workspace = true
 bouncycastle-core.workspace = true
 bouncycastle-utils.workspace = true
 
-# bouncycastle-sha2, -sha3 and -rng are dev-dependencies so that the tests, benches and doc examples
+# bouncycastle-sha2, -sha3, -sm3 and -rng are dev-dependencies so that the tests, benches and doc examples
 # here can still exercise HMAC over the library's own hashes; Cargo permits cycles through
 # dev-dependencies.
 # todo -- we're about to change that and move them to their respective crates in the next phase.
```
ounsworth: Note-to-self: remove this entire comment block.

crypto/sha2/src/lib.rs
```diff
@@ -40,43 +40,27 @@
...
+//! 3 message bits `101`:
+//! ```
+//! use bouncycastle_core::traits::Hash;
+//! use bouncycastle_sha2 as sha2;
+//!
+//! let data: &[u8] = b"\x00\x01\x02\x03\x04\x05\x06\x07\x08\x09\x0A\x0B\x0C\x0D\x0E\x0F\xA0";
+//! let mut sha2 = sha2::SHA256::new();
+//! sha2.do_update(&data[..16]);
+//! let output: Vec<u8> = sha2.do_final_partial_bits(data[16], 3).expect("num_partial_bits is in 0..=7");
+//! ```
```
ounsworth, line 56: Note-to-self: build the docs and look that this renders properly.

crypto/sha2/src/lib.rs
```diff
@@ -108,6 +92,43 @@
 //! sha2_resumed.do_update(msg_part2);
 //! let h: Vec<u8> = sha2_resumed.do_final();
 //! ```
+//!
+//! # Memory Usage
+//!
+//! | Object                                                   | Size (bytes) |
+//! |----------------------------------------------------------|--------------|
+//! | `SHA224`, `SHA256`                                       | 112          |
+//! | `SHA384`, `SHA512`, `SHA512t<T>` (incl. `SHA512_224`, `SHA512_256`) | 208 |
```
ounsworth: Clean this row up.

crypto/sha2/src/lib.rs
```diff
@@ -146,16 +174,116 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
 pub type SHA384 = SHA512Internal<SHA384Params>;
 /// Public type for SHA512.
 pub type SHA512 = SHA512Internal<SHA512Params>;
+/// Public type for the SHA-512/t truncating family (FIPS 180-4 s. 5.3.6): SHA-512 with a
+/// t-specific initial hash value, truncated to `T` bits.
```
ounsworth: Add: "This documentation explains how to instantiate the [`SHA512t'] struct from outside the library with a truncation length other than the 224 and 256 specified in FIPS 180-4.

crypto/sha2/src/lib.rs
```diff
@@ -146,16 +174,116 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
+///
+/// `T` may be any truncation the standard defines a hash for -- "any positive integer without a
+/// leading zero such that t < 512, and t is not 384" -- narrowed here to multiples of 8, since the
+/// digest has to be a whole number of bytes. Anything else is a compile error naming the rule it
+/// broke. The initial hash value is produced at compile time by the s. 5.3.6 IV Generation
+/// Function, so a new `T` costs nothing at runtime and needs no table.
```
ounsworth: If the whole point is to expose this publicly, then it's actually broader than what the standard defines, right? This paragraph needs a re-write.

crypto/sha2/src/lib.rs
```diff
@@ -146,16 +174,116 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
+///
+/// ```compile_fail
+/// use bouncycastle_sha2::SHA512t;
+/// // FIPS 180-4 s. 5.3.6: "t is not 384" -- SHA384 is its own algorithm with its own IV.
+/// let _ = SHA512t::<384>::new_allow_unapproved_t();
+/// ```
+///
+/// See [`SHA512_224`] and [`SHA512_256`] for the approved pair, which are aliases of this type and
+/// are additionally the only truncations with an assigned [`AlgorithmOID`] and a `HashFactory`
+/// entry.
```
ounsworth: Put a forward-ref to this section in the top-of-crate docs. Maybe:

```
# SHA512t

See [`SHA512t`] for documentation around defining a custom truncation length of SHA512 other than the [`SHA512_224`] and [`SHA512_256`] defined in FIPS 180-4.
```

ounsworth: Note-to-self: Build the docs and check that this renders nicely.

crypto/sha2/src/lib.rs
ounsworth: Note-to-self: go look at the tests and guard code.
Are we blocking t = 384? Should we? Does that need a unit test?
Are we blocking t not a multiple of 8? Should we? Does that need a unit test?

ounsworth: Question for David:
What is the initial hash value? Is that just a KAT for testing, or is that a parameter for ... something?

crypto/sha2/src/lib.rs
```diff
@@ -146,16 +174,116 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
+///
+/// // An approved truncation: the ordinary constructor.
+/// let digest = SHA512_256::new().hash(b"abc");
+/// assert_eq!(digest.len(), 32);
+///
+/// // SHA512t<256> *is* SHA512_256.
+/// assert_eq!(SHA512t::<256>::new().hash(b"abc"), digest);
+/// ```
+///
+/// # Only `T = 224` and `T = 256` are approved
```
ounsworth: For clarity, I would say "# Only `T = 224` and `T = 256` are approved by FIPS 180-4"

crypto/sha2/src/lib.rs
```diff
@@ -146,16 +174,116 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
+///
+/// FIPS 180-4 s. 5.3.6 approves exactly two truncations, SHA-512/224 and SHA-512/256 ("Other
+/// SHA-512/t hash algorithms with different t values may be specified in [SP 800-107] in the
+/// future as the need arises"). Every other `T` is a well-defined SHA-512/t but not an approved
+/// hash algorithm, so it must not be used where an approved one is required.
```
ounsworth: This seems to contradict the paragraph above that says that t>512, t=384, and any t with a leading zero, and anything not a multiple of 8 are not valid, so I don't think this paragraph can be correct.

crypto/sha2/src/lib.rs
```diff
@@ -146,16 +174,116 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
+///
+/// That distinction is enforced rather than merely documented, in the same shape as
+/// `ElectronicCodeBook::ENCRYPTION_APPROVED` in the cipher traits: the unapproved truncations carry
+/// [`SHA512tParams::FIPS_APPROVED`]` == false`, and
+/// [`SHA512Internal::new`](sha512::SHA512Internal::new) checks it in an inline `const`. Building
+/// one the ordinary way -- including through `Default`, and so through any generic code that
+/// requires it -- is therefore a compile error at the call site, and
+/// [`SHA512Internal::new_allow_unapproved_t`](sha512::SHA512Internal::new_allow_unapproved_t) is
+/// the way to say you meant it:
```
ounsworth: This paragraph seems weird to me, especially the comparison with ElectronicCodeBook::ENCRYPTION_APPROVED, which is a completely different crate.

I think there is useful content in this paragraph, but it should be distilled down a bit.

crypto/sha2/src/lib.rs
```diff
@@ -245,7 +370,112 @@ impl AlgorithmOID for SHA512 {
...
+const _: () = assert!(matches!(SHA512tParams::<224>::ALG_NAME_STR.as_bytes(), b"SHA512/224"));
+const _: () = assert!(matches!(SHA512tParams::<256>::ALG_NAME_STR.as_bytes(), b"SHA512/256"));
+const _: () = assert!(matches!(SHA512_224_NAME.as_bytes(), b"SHA512/224"));
+const _: () = assert!(matches!(SHA512_256_NAME.as_bytes(), b"SHA512/256"));
+const _: () = assert!(SHA512tParams::<224>::OUTPUT_LEN == 28);
+const _: () = assert!(SHA512tParams::<256>::OUTPUT_LEN == 32);
+const _: () = assert!(SHA512tParams::<224>::FIPS_APPROVED);
+const _: () = assert!(SHA512tParams::<256>::FIPS_APPROVED);
+const _: () = assert!(!SHA512tParams::<8>::FIPS_APPROVED);
+const _: () = assert!(!SHA512tParams::<504>::FIPS_APPROVED);
```
ounsworth: This is fundamentally test code (but done in a way that will cause the build to fail). Can it be moved to `tests/`?

crypto/sha2/src/lib.rs
```diff
@@ -108,6 +92,43 @@
...
+//!   SHA-512/224 and SHA-512/256 offer 112 and 128 bits (SP 800-107r1, Table 1 (§4.2)). More
+//!   generally SHA-512/t offers t/2 bits, which is what [`SHA512t`]'s `MAX_SECURITY_STRENGTH`
+//!   reports, rounded down to a modelled level.
+//! * **Only two SHA-512/t truncations are approved.** [`SHA512t`] is generic over `T`, but FIPS
+//!   180-4 s. 5.3.6 approves only t = 224 and t = 256. Any other `T` is a well-defined hash that
+//!   is nonetheless unapproved, and has to be constructed through
+//!   [`SHA512Internal::new_allow_unapproved_t`](sha512::SHA512Internal::new_allow_unapproved_t)
+//!   rather than `new()`; see [`SHA512t`] for the reasoning and the compile-time gate. Small `t`
+//!   is also simply weak -- SHA-512/8 has a one-byte digest -- and carries a
+//!   `SecurityStrength::None`.
```
ounsworth: See my comments below. I don't think that "FIPS-approved" is a "Security Consideration". FIPS != Security.

As I note below, you're sneaking the beginnings of fips_mode architecture into a PR about truncated SHA512. I don't think it's necessary to implement SHA512t, and I suggest removing it so that we can deal with how to deal with fips_mode as a cohesive architecture topic across the whole library.

crypto/sha2/src/lib.rs
```diff
@@ -146,16 +174,116 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
+///
+/// ```
+/// use bouncycastle_core::traits::{Algorithm, Hash};
+/// use bouncycastle_sha2::{SHA512t, SHA512tParams};
+///
+/// assert!(!SHA512tParams::<96>::FIPS_APPROVED);
+/// let digest = SHA512t::<96>::new_allow_unapproved_t().hash(b"");
+/// assert_eq!(digest.len(), 12);
+/// assert_eq!(<SHA512t<96> as Algorithm>::ALG_NAME, "SHA512/96");
+/// ```
```
ounsworth, line 243: I don't love the `::new_unapproved()` pattern.

I see that this is starting to wade into the problem of "Things that are physically possible, but not allowed in FIPS mode". There will be lots of things of this shape across the library -- for example the escape hatches in KeyMaterial that let you coerce security strengths, stuff in the DRBG crate, etc.

We're going to have to do a big architecture across the library for this so that it feels cohesive. My intuition would be to drive it from a workspace-level cargo feature

```
fips_mode = true
```

and then you put guards on stuff:

```
#[cfg(fips_mode)]
```

like make the SHA512tParams trait sealed when building in FIPS mode:


We'll also need to come up with a docs template for how we communicate that a function or struct has different behaviour in FIPS and non-FIPS modes. I want to look into whether it's possible to have sections of docs get "disabled" based on a cargo feature, such as:

```
#[cfg(!fips_mode)]
{
//! You see these docs in a normal build.
}
#[cfg(fips_mode)]
{
//! And these docs when
//! building in fips_mode.
}
```

and that way the public docs on crates.io aren't polluted by any of the fips_mode stuff -- and people who are working against the fips build get docs specific to that build. But that's an idea in it's infancy and needs some exploration.

Anyway, the point is that I think you should just delete the FIPS_APPROVED flag from [SHA512tParams] -- (which maybe means that the entire SHA512tParams trait can be deleted since it's now 100% duplicate with the Algorithm trait?), and also delete `new_allow_unapproved_t()`.


MOREOVER, where is `::new_allow_unapproved_t()` even defined!? There are 12 CTRL+F hits for that in this file, and all of them are comments. What?

ounsworth (changes requested):

crypto/sha2/src/lib.rs
dghgit: It's been moved to tests, just to note: CLAUDE.md still says const assertions stay in src. If we want this to happen when the tools are in use, we'll need to revise the entry to catch this case as well.

crypto/sha2/src/lib.rs
dghgit: The non-approved gate has been deleted.

crypto/sha2/src/lib.rs
dghgit: The "initial hash value" is the H(0) value that seeds the compression function - it needs to be distinct for each different value of t.

With 384, in true Rust fashion, it will not even compile.

crypto/sha2/src/lib.rs
dghgit: It's correct, it's more "engineering reality meeting specification" kind of thing. I guess the other thing I should point out is that the reason SHA512/224 and SHA512/256 are approved is that they are the only ones ACVP testing is available for. There's nothing to stop people using SHA512/160 or SHA512/320 in unapproved operations, anymore than there is anything stopping the CAVP adding them to ACVP and them suddenly becoming approved.

dghgit:
I think everything in the new comment list has been addressed with specific text answers where the 2 new commits didn't seem enough.

crypto/sha2/src/lib.rs
```diff
@@ -174,68 +177,73 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
+/// hash value H(0): the eight 64-bit words the compression function starts from. FIPS 180-4
+/// s. 5.3.6 requires this of the family -- "Each hash function requires a distinct initial hash
+/// value" -- so where s. 5.3.5 fixes H(0) for SHA-512, s. 5.3.6's IV Generation Function derives
+/// a fresh one for each t by hashing the ASCII string "SHA-512/t" with a modified SHA-512. Because
+/// the starting state differs, a SHA-512/t digest is not a prefix of the SHA-512 digest of the
+/// same message, and digests for different t are unrelated to one another rather than prefixes of
+/// a common value. H(0) is not a parameter the caller supplies: [`SHA512tParams`] computes it at
+/// compile time from `T`, so a new `T` costs nothing at runtime and needs no table, and for
+/// t = 224 and t = 256 the result is pinned by `tests/sha512t_h0_tests.rs` against the words
+/// FIPS 180-4 prints in s. 5.3.6.1 and s. 5.3.6.2.
```
ounsworth, line 229: This in an excellent addition, and actually really clarifies my own understanding of what SHA512t is 👍

crypto/sha2/src/lib.rs
```diff
@@ -174,68 +177,73 @@ pub type SHA256 = SHA256Internal<SHA256Params>;
...
-/// broke. The initial hash value is produced at compile time by the s. 5.3.6 IV Generation
-/// Function, so a new `T` costs nothing at runtime and needs no table.
+/// # Which `T` are accepted
 ///
+/// FIPS 180-4 s. 5.3.6 defines SHA-512/t for every "positive integer without a leading zero such
+/// that t < 512, and t is not 384", and then names two members of that family, SHA-512/224 and
+/// SHA-512/256, as approved hash algorithms. This type implements the family as the section
+/// defines it, with one narrowing of this crate's own: `T` must be a multiple of 8, because
+/// [`Hash`] produces whole bytes. So `T` may be any multiple of 8 from 8 to 504 other than 384;
+/// t = 384 is excluded by the standard because SHA-384 is its own algorithm (s. 5.3.4) with an
```
ounsworth: It might be worth mentioning that SHA384 is already truncated SHA512.
Otherwise, this paragraph sortof doesn't make sense why 384 would be forbidden but not 256 or 224.

crypto/sha2/src/lib.rs
```diff
@@ -384,26 +382,13 @@ impl SHA512InitValue for SHA512Params {
 /// * FIPS 180-4 s. 5.3.6's own rule, "t is any positive integer without a leading zero such that
 ///   t < 512, and t is not 384";
 /// * this crate's additional requirement that `T` be a multiple of 8, since the digest has to be a
-///   whole number of bytes. See [`sha512::sha512t_h0`] for why.
+///   whole number of bytes. See `sha512t_h0` in `sha512.rs` for why.
```
ounsworth: This is a bit weird because you have a docstring on a `pub struct` (which will thus appear in the public crate docs), referencing `sha512t_h0`, which is `pub(crate)`, and therefore to follow this reference, you have to leave the crate docs and go to the source code. 

That's a weird disconnect.

I think the important question here is: "Is the "See ___ for why" intended for external users of the library (in which case the entire reference stack should be in the public dacs), or for maintainers of the library (in which case none of this should be in the public docs)?

I think my vote is that

> `T` be a multiple of 8, since the digest has to be a  whole number of bytes.

is already clear enough, it doesn't really need a "See ___ for more" reference.

ounsworth:
I re-reviewed the SHA512t stuff. Much improved. I left a couple of minor comments, but I think it's 98% there.

crypto/core/src/traits.rs
```diff
@@ -14,8 +14,67 @@ use crate::key_material::KeyType;
 
 /// The basic functions of an Authenticated Encryption with Addititional Data cipher.
 pub trait AEADCipher<const KEY_LEN: usize, const NONCE_LEN: usize, const TAG_LEN: usize>:
```
ounsworth, line 18: Everything else is split into separate Encryptor and Decryptor traits. Why not this?

crypto/core/src/traits.rs
```diff
@@ -14,8 +14,67 @@ use crate::key_material::KeyType;
...
-    SymmetricCipher<KEY_LEN, NONCE_LEN> + Sized
+    Algorithm + Sized
 {
+    #[cfg(feature = "std")]
+    /// A one-shot API to encrypt some plaintext with the given key, with no additional
+    /// authenticated data.
+    ///
+    /// This and the three that follow were the whole of the former `SymmetricCipher` trait, which
```
ounsworth: Comments like:

> This and the three that follow were the whole of the former `SymmetricCipher` trait,

are a point-in-time comment that really doesn't belong in the source code at all, at certainly not in a public API docstring.

I'm gonna go ahead and remove this.

crypto/core/src/traits.rs
```diff
@@ -14,8 +14,67 @@ use crate::key_material::KeyType;
...
+    /// without `std`.
+    ///
+    /// See the documentation for the underlying implementation for how big the ciphertext buffer
+    /// must be; an AEAD needs room for the tag as well as the data. Returns the generated nonce and
+    /// the number of bytes written.
+    fn encrypt_out(
+        key: &KeyMaterial<KEY_LEN>,
+        plaintext: &[u8],
+        ciphertext: &mut [u8],
+    ) -> Result<([u8; NONCE_LEN], usize), SymmetricCipherError>;
```
ounsworth, line 42: Design question: do we want to allow for the input `plaintext` buffer and the output `ciphertext` buffer to be the same buffer?
That's the way that BlockCipherEncryptor below works, and I kinda like it.
I suspect that for embedded systems dealing with large data, that will be a desirable mode.

I think would need to separate out the tag into a separate return value so that the length of the plaintext and ciphertext match (which might just be a good idea anyway).

crypto/core/src/traits.rs
```diff
@@ -95,73 +154,254 @@ pub trait AlgorithmOID {
...
+/// Implementors are expected to hold that key schedule in a zeroize-on-drop wrapper
+/// (`bouncycastle_utils::secret::Secret`), so it is scrubbed when the value is dropped.
+///
+/// # Why the block methods are infallible
+///
+/// Every length here is fixed by a type, and a constructed value is always ready to use, so there
+/// is nothing a caller can get wrong once [`ElectronicCodeBook::new`] has returned. Only `new` can
+/// fail, and only because of the key.
+pub trait ElectronicCodeBook<const KEY_LEN: usize, const BLOCK_LEN: usize>:
+    Algorithm + Sized
```
ounsworth, line 350: Since this trait is not intended for external users, but for library maintainers building CBC / GCM / etc modes, I suggest that we move this trait out of `bouncycastle_core` and hide it in the crate `bouncycastle_cipher_utils`.

Aside: I suggest taking the `modes` and `padding` crates that you have now and collapsing them into a single `bouncycastle_cipher_utils`.

crypto/core/src/traits.rs
```diff
@@ -95,73 +154,254 @@ pub trait AlgorithmOID {
...
+    /// results. `TestFrameworkElectronicCodeBook` pins that.
+    ///
+    /// Modes whose structure is parallel -- CBC decryption, CFB decryption, CTR -- should prefer
+    /// this. CBC and CFB *encryption* cannot use it: each input block depends on the previous
+    /// output.
+    fn encrypt_2blocks(&self, blocks: &mut [[u8; BLOCK_LEN]; 2]) {
+        let [a, b] = blocks;
+        self.encrypt_block(a);
+        self.encrypt_block(b);
+    }
```
ounsworth, line 382: Question because I don't know the answer: I know that our AES implementation will support this two-blocks-at-a-time mode, but will that be true of all block ciphers that we need to implement, or will having this as a mandatory function on the ECB trait become a problem in the future?

Ditto for the `_4blocks` below.

crypto/core/src/traits.rs
```diff
@@ -544,6 +795,34 @@ pub trait MAC: Sized {
     fn max_security_strength(&self) -> SecurityStrength;
 }
 
+/// A block padding scheme, used to extend arbitrary-length data to a whole number of blocks so that it
+/// can be processed by a [`BlockCipherEncryptor`]. Implementations are pure functions of the block
+/// contents: no key, no state.
+///
+/// Only the final, partial block of a message is ever padded; the padding layer sitting between the
+/// caller and the block cipher is responsible for routing whole blocks straight through.
+pub trait Padding<const BLOCK_LEN: usize> {
```
ounsworth: Since this is not intended to be used by external users, I suggest burying this in the `bouncycastle-cipher-utils` crate.

ounsworth: Also, this is called "Padding", but is in fact specific to block ciphers. So that's gonna be a problem when we get to RSA and other non-block-cipher things that also have a component called "padding". Tucking this under a "bouncycastle-cipher-utils` crate / namespace also solves that problem.

crypto/core/src/traits.rs
```diff
@@ -95,73 +154,254 @@ pub trait AlgorithmOID {
...
+        &mut self,
+        blocks: &mut [[u8; BLOCK_LEN]],
+    ) -> Result<(), SymmetricCipherError>;
+
+    /// Streaming: decrypts `LEN` bytes, a whole number of blocks, in place. `LEN % BLOCK_LEN == 0`
+    /// is checked at compile time, exactly as for [`BlockCipherEncryptor::do_encrypt`].
+    fn do_decrypt<const LEN: usize>(
+        &mut self,
+        data: &mut [u8; LEN],
+    ) -> Result<(), SymmetricCipherError> {
```
ounsworth: This feels wrong to me. We're requiring applications to know *at compile time* how big the data is that they are encrypting / decrypting?

> Error: email attatchments must be exactly 25 mb because that's the size we compiled our crypto library for

I would think that the length of data is the one thing that, by definition, needs to be a runtime variable, no?

crypto/core/src/traits.rs
ounsworth: For all ciphers that act in-place, it's probably worth a docs note that if you want to preserve your plaintext data, then you need to clone it yourself.

crypto/core/src/traits.rs
```diff
@@ -95,73 +154,254 @@ pub trait AlgorithmOID {
...
-    SymmetricCipher<KEY_LEN, INIT_DATA_LEN> + Sized
+///
+/// # Everything is in place
+///
+/// Every data method here transforms its buffer in place: the plaintext goes in, the ciphertext
+/// comes out in the same bytes. A block cipher mode never changes the length of its data, so a
+/// separate output buffer would only ever be a copy, and a copy of plaintext is one more thing to
+/// scrub. Callers that need to keep the plaintext copy it first.
+///
+/// # Lengths are checked at compile time
```
ounsworth, line 229: As I noted above in the decryptor, this doesn't make sense to me as a design pattern. How is it even possible for an application to know the size if its data buffer until runtime? Isn't that by definition something that it can't know until it's read the data off a network socket or from a file?

crypto/core/src/traits.rs
```diff
@@ -95,73 +154,254 @@ pub trait AlgorithmOID {
...
+    /// permutation processes several blocks at once (CBC decryption, CTR) chunks the slice itself.
+    /// Callers should normally use the flat [`BlockCipherEncryptor::do_encrypt`] instead.
+    fn do_encrypt_blocks(
         &mut self,
-        ciphertext: &[u8; BLOCK_LEN],
-    ) -> Result<[u8; BLOCK_LEN], SymmetricCipherError>;
-    /// Decrypts the final block of ciphertext and writes the plaintext to the provided buffer.
-    fn do_decrypt_final_out(
+        blocks: &mut [[u8; BLOCK_LEN]],
+    ) -> Result<(), SymmetricCipherError>;
```
ounsworth: Everywhere else in the library that we write to a provided output buffer, we return `usize`: the number of bytes written. Please stick to the consistent pattern.

crypto/core/src/traits.rs
```diff
@@ -95,73 +154,254 @@ pub trait AlgorithmOID {
...
+/// separate output buffer would only ever be a copy, and a copy of plaintext is one more thing to
+/// scrub. Callers that need to keep the plaintext copy it first.
+///
+/// # Lengths are checked at compile time
+///
+/// Every buffer is a `[u8; LEN]`, and `LEN % BLOCK_LEN == 0` is checked by an inline `const`
+/// assertion when the method is instantiated: a misaligned length is a compile error at the call
+/// site, not a runtime `Err`, which is why there is no length variant of [`SymmetricCipherError`]
+/// here. Data whose length is only known at run time is fed in block by block, or through the
+/// padding layer.
```
ounsworth, line 235: Question just for my understanding: is BlockCipherEncryptor / BlockCipherDecryptor something that we expect external users to use, like is this the public API that they'll use to interact with AES_CBC?
Or is there some other trait that would be the public interface to AES_CBC?


If this is the public interface to AES_CBC, then expecting the data to have `LEN % BLOCK_LEN == 0` is unreasonable because this primitive will internally do the padding (which means you need to account for the ciphertext potentially being a bit larger than the plaintext.

One design pattern that I haven't explored yet, but have in the back of my mind is that we could return an error of type `IncorrectBufferSize(usize)` where we return the length that that buffer should have been so that they can size it correctly and try again. I think that pattern is common in openssl where users will just do a call with a 0-size buffer that will fail, but it'll tell them how big to size it. I haven't used that pattern anywhere in the library yet, but Jason and I have discussed that it's probably the way to go in order to get the Hex and Base64 crates to play nicely in no_std, no alloc. I'm ok with starting to introduce that pattern.

crypto/core/src/traits.rs
```diff
@@ -875,55 +1154,118 @@ pub trait Signer<SK: SignaturePrivateKey<SK_LEN>, const SK_LEN: usize, const SIG
...
+/// check to check. How the keystream is produced internally -- in 64-byte blocks, in words, a bit
+/// at a time -- is the cipher's business; it buffers any unused keystream between calls so that
+/// the caller's chunking is never visible in the output.
+///
+/// # Why the data methods still return `Result`
+///
+/// Nothing about the buffer can go wrong, and a constructed value is always ready to use. The
+/// `Result` is for the per-initialization data limit most stream ciphers have: a counter-driven
+/// keystream must refuse to run past the point where its counter would wrap and the keystream
+/// repeat, and a streaming API cannot check that any earlier than the call that would cross it.
```
ounsworth, line 1247: I think this comment about the fallibility of functions should go on each function so that it shows up in in-IDE tooltip popups, not on the trait (which tends to get lost in IDEs).

crypto/core/src/traits.rs
```diff
@@ -875,55 +1154,118 @@ pub trait Signer<SK: SignaturePrivateKey<SK_LEN>, const SK_LEN: usize, const SIG
...
+    fn do_encrypt_init_rng(
+        key: &KeyMaterial<KEY_LEN>,
+        rng: &mut dyn RNG,
+    ) -> Result<(Self, [u8; INIT_DATA_LEN]), SymmetricCipherError>;
+
+    /// Streaming: encrypts `data`, of any length, in place. A sequence of calls is equivalent to
+    /// one call over the concatenation, whatever the chunking.
+    ///
+    /// This is the only method an implementor writes besides the two `_init` constructors.
+    fn do_encrypt(&mut self, data: &mut [u8]) -> Result<(), SymmetricCipherError>;
```
ounsworth: For consistency with the rest of the library, this should return `usize`: the number of bytes written, which usually will be exactly `data.len()`, but we should still return it.

crypto/core/src/traits.rs
```diff
@@ -875,55 +1154,118 @@ pub trait Signer<SK: SignaturePrivateKey<SK_LEN>, const SK_LEN: usize, const SIG
...
+        let (mut enc, init_data) = Self::do_encrypt_init(key)?;
+        enc.do_encrypt(data)?;
+        Ok(init_data)
+    }
+    /// As [`StreamCipherEncryptor::encrypt`], but sources randomness from the provided RNG.
+    fn encrypt_rng(
+        key: &KeyMaterial<KEY_LEN>,
+        rng: &mut dyn RNG,
+        data: &mut [u8],
+    ) -> Result<[u8; INIT_DATA_LEN], SymmetricCipherError> {
```
ounsworth: For consistency with the rest of the library, this should return `usize`: the number of bytes written, which usually will be exactly `data.len()`, but we should still return it.

crypto/core/src/traits.rs
```diff
@@ -875,55 +1154,118 @@ pub trait Signer<SK: SignaturePrivateKey<SK_LEN>, const SK_LEN: usize, const SIG
...
+    /// one call over the concatenation, whatever the chunking.
+    ///
+    /// This is the only method an implementor writes besides the two `_init` constructors.
+    fn do_encrypt(&mut self, data: &mut [u8]) -> Result<(), SymmetricCipherError>;
+
+    /// One-shot: encrypts `data` in place under a fresh init, and returns the generated init data.
+    fn encrypt(
+        key: &KeyMaterial<KEY_LEN>,
+        data: &mut [u8],
+    ) -> Result<[u8; INIT_DATA_LEN], SymmetricCipherError> {
```
ounsworth: For consistency with the rest of the library, this should return `usize`: the number of bytes written, which usually will be exactly `data.len()`, but we should still return it.

crypto/core/src/traits.rs
```diff
@@ -875,55 +1154,118 @@ pub trait Signer<SK: SignaturePrivateKey<SK_LEN>, const SK_LEN: usize, const SIG
...
+    /// one call over the concatenation, whatever the chunking, exactly as for
+    /// [`StreamCipherEncryptor::do_encrypt`].
+    fn do_decrypt(&mut self, data: &mut [u8]) -> Result<(), SymmetricCipherError>;
+
+    /// One-shot: decrypts `data` in place from the given init data.
+    fn decrypt(
+        key: &KeyMaterial<KEY_LEN>,
+        init_data: &[u8; INIT_DATA_LEN],
+        data: &mut [u8],
+    ) -> Result<(), SymmetricCipherError> {
```
ounsworth: For consistency with the rest of the library, this should return `usize`: the number of bytes written, which usually will be exactly `data.len()`, but we should still return it.

crypto/core/src/traits.rs
```diff
@@ -875,55 +1154,118 @@ pub trait Signer<SK: SignaturePrivateKey<SK_LEN>, const SK_LEN: usize, const SIG
...
-    fn do_stream_decrypt_block_out<const BLOCK_LEN: usize>(
-        &mut self,
-        ciphertext: &[u8; BLOCK_LEN],
-        plaintext: &mut [u8; BLOCK_LEN],
-    ) -> Result<usize, SymmetricCipherError>;
+
+    /// Streaming: decrypts `data`, of any length, in place. A sequence of calls is equivalent to
+    /// one call over the concatenation, whatever the chunking, exactly as for
+    /// [`StreamCipherEncryptor::do_encrypt`].
+    fn do_decrypt(&mut self, data: &mut [u8]) -> Result<(), SymmetricCipherError>;
```
ounsworth: For consistency with the rest of the library, this should return `usize`: the number of bytes written, which usually will be exactly `data.len()`, but we should still return it.

ounsworth (changes requested):

crypto/core/src/traits.rs
dghgit: It'll vary from 1 to 4. The highest throughput AES instructions are currently 4 lane (_4blocks). The circuit for SM4 did actually start at 8 lane, but I managed to talk the LLM down to 4 as well. At the moment I'm not aware of anything that goes above 4 lane.

crypto/core/src/traits.rs
dghgit: Note sure with this one, we really need to add a second block cipher padding to see how it plays out. Note decisions around whether to follow the utility crate model used with the message digests will also affect this. I'd suggest adding some additional paddings first (maybe TBC (trailing bit complement - often referred to in NIST documents, although rarely seen...) and one of the ISO ones) and looking at it again when it's possible to do a "grand refactor". Agree about the naming, now called BlockCipherPadding.

crypto/core/src/traits.rs
dghgit: This PR is about dealing with block and stream ciphers. #120 is where the AEAD cipher split takes place, including AEAD splitting here would have meant also including Ascon, GCM, and CCM, which seemed a bit much, so I figured we'd leave it broken and fix it later.

Experience has taught me that editing when you can't validate the changes is not a good use of anyone's time.

crypto/core/src/traits.rs
dghgit: Wrong trait for the question, it provides a building block for other things, it is also something people would use if they can meet the constraints required though, the SimpleCipherEncryptor/Decryptor are the "unifiers" in this case, they're the ones that manage modes and padding in-line.

crypto/core/src/traits.rs
dghgit: I don't understand. I think you're talking about SimpleCipher...., but CBC is a block cipher mode which requires block aligned data. There is an IETF standard which defines the CBC OID as representing CBC with PKCS#7Padding, and it's true for many IETF standards, but it's not actually correct. CBC is simply an overlay on ECB, padding is defined separately.

crypto/core/src/traits.rs
dghgit: The other thing is there's nothing to stop someone providing their own padding mechanism - even FIPS doesn't forbid that. For example some people simply pad with Zeros. While we support that in bc-java, I wouldn't support it here as it's a terrible idea, on the other hand if someone was willing to do the work we should probably allow it as they probably do not have a choice in the matter.

crypto/core/src/traits.rs
dghgit: Done.

crypto/core/src/traits.rs
dghgit: Done

crypto/core/src/traits.rs
dghgit: It's tempting to ask what an IDE is, but sure... I'd do this when your merging though, just to make sure the edits have the desired effect.

crypto/core/src/traits.rs
dghgit: Done.

crypto/core/src/traits.rs
dghgit: Done.

crypto/core/src/traits.rs
dghgit: Done

crypto/core/src/traits.rs
dghgit: I'd wait till GCM, CCM and co have landed - modes and padding are distinct, that said padding only appears with block ciphers and block-cipher modes which also behave as block ciphers (really just CBC). The OpenSSL list has ASX (which is AES only) and CTS (which can be used with any block cipher that supports CBC). In a sense you could say paddings here are a subset of modes (Padding renamed BlockCipherPadding). I think it's really a "grand refactor" item, and we do need a couple more paddings for block ciphers.

crypto/core/src/traits.rs
dghgit: I've captured this in QUALITY_AND_STYLE.md and done it.

crypto/core/src/traits.rs
dghgit: Just ignore this one see #120, what you're looking at here is what is currently on release/0.1.3alpha.

crypto/core/src/traits.rs
dghgit: Again #120, I actually like the return tag separate option and that's been preserved - it's not something you see often though. Usually the tag always appears at the end of the cipher text, however there are a number of situations, including EnvelopedAuthenticatedData in CMS where the protocols specify the tag is to be separate. This can be handy as AEAD ciphers (at least to date) don't change the size of the cipher-text/plain-text outside, so it does offer the option for in-place encryption/decryption, we should follow this more closely in #120.

crypto/sha2/src/lib.rs
dghgit: Done.

crypto/sha2/src/lib.rs
dghgit: Done.

crypto/core/src/traits.rs
ounsworth: Then why is this PR changing the entire body of the AEADCipher trait? The whole thing shows up as green in the diff, which is why I feel like it's in-scope to comment on as part of this PR. Something's not adding up here.

crypto/core/src/traits.rs
ounsworth: Ok.

But again, it's showing up as green in the diff. In the future, to avoid wasting reviewer's time, if something is not meant to be included and reviewed as part of a PR, then don't change it so that it shows up green in the diff.

crypto/core/src/traits.rs
ounsworth: (ditto above)

crypto/core/src/traits.rs
ounsworth: I see.

I think the point here is that if I have to ask, then the docs weren't clear enough.

I think what you're telling me is that the BlockCipherEncryptor trait is not intended to be part of the public API of the library, but rather it's an internal building-block trait?

I suggest that we put some thought into how to communicate to both external users and library maintainers what is public API surface, and what is internal building blocks. One pseudo-objective of mine for the bouncycastle-core crate is that it *could* be used as a common API across crypto libraries -- ie you *could* wrap RustCrypto or Graviola in these traits, so I've been trying to keep implementation-specific things out (though, I think that's gonna get harder and harder as time goes on).

So, option for communicating Public API vs Internal API:

1. Take Internal API traits out of `bouncycastle-core` and move them to a suitable utility crate (in this case, bouncycastle-cipher-utils or equiv.
2. Leave Internal API Traits here, but come up with a docstring convention for tagging them. See below.
3. Demote them down one level to a new module `bouncycastle-core::internal::*`.
4. Something else?


Proposal for how we could tag internal traits (option 2 above): put a `///` public docstring just saying "Internal, move along", and then put a `//` comment block explaining what it does for library maintainers.

```rust
/// Internal API used between bouncycastle-* crates, but not intended to be part of the library's public API surface.
// In a non-public docstring, we put the full explanation for library maintainers.
pub trait NotReallyPartOfPublicAPI { }
```

crypto/core/src/traits.rs
ounsworth: I understand that cipher padding and cipher modes are distinct. I understand that some ciphers will use one but not the other, and some ciphers will use both, and some ciphers will use neither.

I'm asking a different question, which is entirely to do with packaging and trying to reduce the total number of crates, where possible. Let me try to ask it in a different way:

Crates are the fundamental unit of publishing in Rust. When we eventually put bc-rust on https://crates.io, each crate becomes a separately published library. So there is a (small, but non-zero) overhead cost to each additional crate.

I understand fully why CipherModes and CipherPadding needs to be separate from the AES crate: because it will be shared code between AES, TDES, Camelia, etc.

The question I'm asking here is whether there is a structural reason why `CipherModes` and `CipherPadding` _need_ to be separate crates from each other?
Is there a reason that we couldn't make one `bouncycastle-cipher-utils` and take the two current things and tuck them into sub-modules:
* `bouncycastle-cipher-utils::modes`
* `bouncycastle-cipher-utils::padding`
?

crypto/core/src/traits.rs
ounsworth: I don't believe that you answered my question.

Let me try asking it a different way.

Besides AES, what block ciphers will we implement in the library in the future? I assume at least TDES, Aria, Camellia, BlowFish, etc. 


Traits in Rust cannot have optional functions. If something impl's a trait, then it must impl all functions of that trait.

**Question**: do we know that _every_ block cipher has a 2block and 4block implementation? If not, then isn't this setting us up for this trait being unimplementable on some future primitives?

I haven't yet got to look at how you're using these functions in the Modes crate, but if the idea of a trait in Rust is to describe a unit of behaviour, and single-block and multi-block are different behaviours where some ciphers may not implement all behaviours, then maybe we should split this into three different traits: `ElectronicCodeBook`, `ElectronicCodeBook2Block`, and `ElectronicCodeBook4Block`. I suspect that would still allow you to build your Modes crate to take advantage of the higher lanes, when available, and fall back to the lower-lane traits otherwise; and do it all through compile time magic based on which traits a given primitive impls.

crypto/core/src/traits.rs
ounsworth: Just because you don't use an IDE anymore doesn't mean that our users won't ;)

Let's leave this comment open and unresolved then.

crypto/core/src/traits.rs
```diff
@@ -989,53 +1348,416 @@ pub trait SuspendableKeyed<const SERIALIZED_STATE_LEN: usize>: Sized {
...
+///
+/// Init data (an IV or nonce) is generated by the constructor and returned, never supplied, for
+/// the same reason as in [`BlockCipherEncryptor`]. Everything is `no_std`-friendly except the
+/// allocating [`encrypt`](Self::encrypt), which sits behind the `std` feature.
+///
+/// The one-shots [`encrypt_out`](Self::encrypt_out) and [`encrypt_out_rng`](Self::encrypt_out_rng)
+/// are provided over the streaming methods. An implementor writes only the two `_init`
+/// constructors, [`update_out_len`](Self::update_out_len), [`do_update_out`](Self::do_update_out),
+/// [`do_final`](Self::do_final) and [`encrypt_out_len`](Self::encrypt_out_len).
+pub trait SimpleCipherEncryptor<
```
ounsworth: If I'm understanding correctly, the intention of this trait is that this is the API that you would use if you just want to encrypt some data, and don't want to know or care whether the thing is a block cipher, stream cipher, AEAD cipher, etc?
That means that this is essentially the primary API for all block ciphers, stream ciphers, and aead ciphers? Is that right?

If so, that could be stated more clearly in the docstring. I might start with something like

```
/// This is a common API implemented by all symmetric ciphers, in addition to their
/// family-specific traits such as [`AEADCipherEncryptor`], [`BlockCipherEncryptor`], etc.
```

ounsworth: Also, are you open to bike-shedding the name? To me, given our naming pattern of `BlockCipherEncrypton`, `StreamCipherEncryptor`, etc, then `SimpleCipherEncryptor` is the encryption API for a Simple Cipher. What is a Simple Cipher? My first reaction to this name was that this is for toy ciphers like Rot13.

If this is intended as an umbrella trait for all families of symmetric ciphers, then maybe it should just be `CipherEncryptor`?

crypto/core/src/traits.rs
```diff
@@ -989,53 +1348,416 @@ pub trait SuspendableKeyed<const SERIALIZED_STATE_LEN: usize>: Sized {
...
+    /// # Errors
+    /// Rejects a key whose [`KeyType`] is not [`KeyType::SymmetricCipherKey`], and one whose
+    /// security strength is below [`Algorithm::MAX_SECURITY_STRENGTH`], both as a
+    /// [`SymmetricCipherError::KeyMaterialError`].
+    fn do_encrypt_init(
+        key: &KeyMaterial<KEY_LEN>,
+    ) -> Result<(Self, [u8; INIT_DATA_LEN]), SymmetricCipherError>;
+
+    /// As [`do_encrypt_init`](Self::do_encrypt_init), but sources randomness from the provided RNG.
+    fn do_encrypt_init_rng(
```
ounsworth, line 1559: We should decide how this is to behave for ciphers that don't require randomness (even if we don't have any of those yet).

For example, if someone where to implement Rot13 under this trait (for educational purposes, say), then how should `Rot13::do_encrypt_init_rng()` behave?

Is it supposed to throw an `unimplemented!()` panic? Or throw some type of `SymmetricCipherError`? Or just gracefully work as an alias of `do_encrypt_init()`? (I vote for the latter)

We should decide and document that in the function contract here.

crypto/core/src/traits.rs
```diff
@@ -989,53 +1348,416 @@ pub trait SuspendableKeyed<const SERIALIZED_STATE_LEN: usize>: Sized {
...
+        key: &KeyMaterial<KEY_LEN>,
+        rng: &mut dyn RNG,
+    ) -> Result<(Self, [u8; INIT_DATA_LEN]), SymmetricCipherError>;
+
+    /// The exact number of bytes the next [`do_update_out`](Self::do_update_out) will write if
+    /// given `input_len` more bytes of plaintext. Depends on what is already buffered.
+    fn update_out_len(&self, input_len: usize) -> usize;
```
ounsworth, line 1580: I don't understand the purpose of this, which I think means that it needs more words. It is to help the user size the `ciphertext` buffer that they will hand in to `do_update_out()`?

If so, I would suggest a different design pattern (the same one that I believe openssl uses), which is that if `do_update_out()` is given too small an output buffer, it should fail with a `SymmetricCipherError::OutputBufferTooSmall(usize)` and return the minimum buffer size for a retry of this call to succeed.

(if we decide to go with that pattern here, then we should open a github ticket to apply that behaviour consistently across all `_out()` functions across the whole library.

ounsworth: Below you say that it produces the error `SymmetricCipherError::IncorrectOutputBufferLength` if the provided buffer is too small.

Is providing an output buffer that's too large ever a problem? I think not. I think that's why our global pattern is to `-> Result(usize, _)` for all `_out()` functions, so that we can gracefully handle oversized buffers.

In that case, I suggest renaming the error to `SymmetricCipherError::OutputBufferTooSmall(usize)`.

crypto/core/src/traits.rs
```diff
@@ -989,53 +1348,416 @@ pub trait SuspendableKeyed<const SERIALIZED_STATE_LEN: usize>: Sized {
...
+        let (buffer, out_len) = self.do_final()?;
+        *ciphertext = buffer;
+        Ok(out_len)
+    }
+
+    /// The exact ciphertext length for a `plaintext_len`-byte plaintext that the cipher accepts,
+    /// i.e. the buffer [`encrypt_out`](Self::encrypt_out) requires and the number of bytes it
+    /// writes. (A length the cipher rejects -- unaligned data under a scheme that adds no padding --
+    /// fails in [`do_final`](Self::do_final) instead.)
+    fn encrypt_out_len(plaintext_len: usize) -> usize;
```
ounsworth, line 1620: Same comment as on `update_out_len()` above.

I would like @jjkurczak 's input on this, because he's been thinking about the analogous problem for making Hex and Base64 work in a no_std, no alloc env.

crypto/core/src/traits.rs
```diff
@@ -989,53 +1348,416 @@ pub trait SuspendableKeyed<const SERIALIZED_STATE_LEN: usize>: Sized {
...
+///
+/// Note that both traits then offer `do_encrypt_init` and `do_encrypt_init_rng` with identical
+/// signatures. Where both are in scope, a call needs qualifying --
+/// `<Cfb<..> as StreamCipherEncryptor<..>>::do_encrypt_init(&key)` -- though either resolves to the
+/// same function.
+impl<T, const KEY_LEN: usize, const INIT_DATA_LEN: usize>
+    SimpleCipherEncryptor<KEY_LEN, INIT_DATA_LEN, 0> for T
+where
+    T: StreamCipherEncryptor<KEY_LEN, INIT_DATA_LEN>,
+{
```
ounsworth, line 1708: This is not a trait, it's an impl.

This file, `traits.rs` is likely to get quite large over time as it is, I suggest thet we keep it clean and only contain traits. Maybe we need to start an `impls.rs` or something?

(as I write this, I'm aware that this file also contains the SecurityStrength enum. I've opened #148 to do that refactor)

ounsworth (changes requested):

crypto/core/src/traits.rs
dghgit: Yes, it needs a better name, its not an AEAD cipher though, they're quite different, the class is kind of the swiss army knife simple cipher (so no authenticated encryption, block with padding, or stream, so arbitrary data length).

Originally these were SymmetricCipherEncryptor/Decryptor... but that didn't really work as it also seemed to imply AEAD, which are symmetric, but authenticated. Maybe we should have a competition.

Maybe SymmetricCipher... really is the better name.

crypto/core/src/traits.rs
dghgit: Look, "security best practice" would suggest throwing unimplemented!() as someone calling it for rot13 would really be giving away that they didn't know what they were doing and needed to be alerted to that (usage of rot13 not withstanding).

crypto/core/src/traits.rs
dghgit: Sure. The JCE actually takes this approach with doFinal() as well, although it also supports the output_len function to provide people with an estimate of what might come back on an update/doFinal. One thing I'm not sure about is the recovery approach does mean not changing the cipher state, I'd guess that would work here though, there's one way to find out. I'll try it.

crypto/core/src/traits.rs
dghgit: No idea on that one. It get's deleted in #120 so I don't think it matters what git thinks is going on.

crypto/core/src/traits.rs
dghgit: As I said, no idea, green means new code usually. It may have been sucked in from some other branch.

crypto/core/src/traits.rs
dghgit: Maybe bouncycastle-symmetric? They're not really utilites. You could even put the algorithms under a name like bouncycastle-symmetric, would that streamline things a bit?

crypto/core/src/traits.rs
dghgit: Splitting won't help as the common trait is used by everything that plays with a block cipher. Every block cipher is capable of processing 4 blocks though, either one at a time, or in the case of a later generation Intel CPU in 4 lanes at once. The tdes, camellia, aria, and sm4 are currently built on 4 lane circuits as well. I think this one is correct in that it will cope with everything we currently know of, so hardware intrinsics, and hand-rolled bitslice implemenations.

It will probably make more sense when you look at the modes crate though.

crypto/core/src/traits.rs
dghgit: This one can be part of the public API, some people will use it, I can guarantee that, I've even seen people try to do this using the JCE... most people will use whatever SimpleCipher... winds up being called though

crypto/core/src/traits.rs
dghgit: impl.rs works for me.

crypto/core/src/traits.rs
dghgit: Okay, the following now works:

```
    let written = match Dec::decrypt_out(key, iv, ciphertext, &mut plaintext) {
        Ok(written) => written,
        Err(SymmetricCipherError::OutputBufferTooSmall(needed)) => {
            println!(
                "  decrypt: output buffer of {} too small, needs {needed} -- growing and retrying",
                plaintext.len()
            );
            plaintext.resize(needed, 0);
            // The buffer check happens before any decryption, so nothing was decrypted and the
            // retry is a fresh, complete decryption of the whole ciphertext.
            Dec::decrypt_out(key, iv, ciphertext, &mut plaintext)?
        }
        Err(other) => return Err(other),
    };
```

crypto/core/src/traits.rs
ounsworth: Ok, but having large blocks of code show up green a PR diff means that a reviewer is going to spend time on it. The shotgun approach that changes all sorts of unrelated stuff makes PRs REALLY time-consuming to review.

So, if this change is not meant to be here, then can you please revert it so that it no longer shows up in the diff?

crypto/core/src/traits.rs
ounsworth: When people talk about "symmetric cryptography" doesn't that usually include things like HMAC?

bouncycastle-cipher ?
That would match all the trait names BlockCipher, StreamCipher, AEADCipher, etc.

crypto/core/src/traits.rs
ounsworth: > The tdes, camellia, aria, and sm4 are currently built on 4 lane circuits as well.

Perfect. That's the answer I was looking for!

And good point that you can always fake a 4-lane mode by just doing the 4 blocks in sequence.

Ok. Good.

crypto/core/src/traits.rs
ounsworth: Ok, then please add that to the docstring. My point is that how this is to behave for a cipher that doesn't require randomness should be documented in the function contract.

My vote is that a cipher that doesn't require randomness should just work (ie this becomes an alias to do_encrypt_init() ), but if you think it should panic, I won't argue the point.

crypto/core/src/traits.rs
ounsworth: I see that you've renamed it to `SymmetricCipherEncryptor` :+1: 

Can you explain why AEAD cipher is fundamentally different?

Let's say I'm an undergrad student playing with crypto for a course project -- or heck, a full-stack web dev who needs to encrypt a file -- there should be an umbrella trait where:

```
Cipher::encrypt( b"Attack at dawn!" );
```

just works. Are AEADs so fundamentally different that they can't encrypt data in a simple way, that you _have_ to know how they work in order to use them?

For this to work, we will have to pack the tag into the ciphertext in a way that may or may not be compatible with a different library, but that's fine.

Am I missing something here?

crypto/core/src/traits.rs
ounsworth: I just want to make sure that we're talking about the same thing.

What I'm asking questions about is 

```
    /// Every buffer is a `[u8; LEN]`, and `LEN % BLOCK_LEN == 0` is checked by an inline `const`
    /// assertion
    /// ...
    fn encrypt<const LEN: usize>(
        key: &KeyMaterial<KEY_LEN>,
        data: &mut [u8; LEN],
    ) -> Result<(usize, [u8; INIT_DATA_LEN]), SymmetricCipherError> {
```
is only useful to users who can provide a LEN at compile time -- ie who are handling fixed-size data.
Those exist of course; as an example, credit card PDUs are fixed-size, but my intuition is that >99% of applications will need to encrypt data whose size is not known at compile time.

My proposal is to remove the LEN compile-time constant to change this from a compile-time check to a runtime check:

```rust
    fn encrypt(
        key: &KeyMaterial<KEY_LEN>,
        data: &mut [u8],
    ) -> Result<(usize, [u8; INIT_DATA_LEN]), SymmetricCipherError> {
```
where it throws a SymmetricCipherError of some type if `data.len() % BLOCK_LEN != 0`.
That's a small API change, but it feels like it would make this API useful to 90% more people than only the users who have fixed-size data to encrypt.

(I suspect what happened here is that the LLM got over-zealous about the style guide point "Do compile-time checks instead of runtime checks where appropriate" -- but in this case it's not appropriate)

crypto/aes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
+//! A constant-time, table-free AES block cipher engine (NIST FIPS 197).
```
ounsworth, line 1: I'm starting to move my review from traits.rs into here.

I thought we had discussed that the `bouncycastle_aes::` namespace is where you would expect to find AES_CBC, AES_CCM, AES_GCM, etc. So this is more than an AES engine, this is where you'll find all AES-based primitives.

That means that this will be the place to document the Usage for all AES modes.

With rust docs, the structure of the code translates directly into the structure of the rendered HTML docs. Right now you have all the sub-modules as private:

```
mod aes;
mod bitslice;
mod cbc;
mod cfb;
mod cfb8;
mod ctr;
mod ecb;
mod padded_mode;
mod round;
mod sbox;
mod schedule;
```

which means that nothing about them shows up in the HTML docs.

I wonder if we want to make the modes files be public modules

```
mod aes;
mod bitslice;
pub mod cbc;
pub mod cfb;
pub mod cfb8;
pub mod ctr;
pub mod ecb;
mod padded_mode;
mod round;
mod sbox;
mod schedule;
```

and then the HTML crate docs will get a sub-page for each one, and then we have a place to put usage examples for each mode instead of cluttering up the main page.

(for example, I did MLDSA that way with `pub mod hash_mldsa` so that it can carry its own docs instead of cluttering up the main page)

crypto/core/src/traits.rs
dghgit: Yes, it was exciting, but we got there... With AEAD, the ciphers carry authentication tags and the tag includes associated data not present in the plain text or cipher stream, so yes, they are fundamentally different. The authentication tags often appear in-line (for example TLS), but are also managed separately (CMS AuthenticatedEnvelopedData for example). From an implementation perspective, the way finalisation is handled is determined by which particular style is in use.

We could write a "Cipher::encrypt( b"Attack at dawn!" )" but you wouldn't do it here, it would be a general purpose utility built on SymmetricCipherEncryptor/Decryptor and AEADCipherEncryptor/Decryptor (see #149), at this stage we're still trying to provide the support for doing something like that.

crypto/aes/src/lib.rs
dghgit: No idea on this one. I'm happy to go with whatever fits better with the "Rust feel", I'd wait till a few more block ciphers have gone in before making a decision on this one. As far as I can tell the modules don't need to be public, and experience has taught me that things should only be public if they have to be, come version 1, if it's public it will be really hard to take it back.

crypto/aes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+//! ## Encrypting and decrypting a single block
+//!
+//! ```
+//! use bouncycastle_aes::AES_128;
+//! use bouncycastle_core::key_material::{KeyMaterial, KeyType};
+//! use bouncycastle_core::traits::ElectronicCodeBook;
+//!
+//! let key = KeyMaterial::<16>::from_bytes_as_type(
+//!     &[0x2b, 0x7e, 0x15, 0x16, 0x28, 0xae, 0xd2, 0xa6,
+//!       0xab, 0xf7, 0x15, 0x88, 0x09, 0xcf, 0x4f, 0x3c],
```
ounsworth: This is maybe paranoia, but everywhere else that I've put a sample key in docs, I've used:

```
&[0x00, 0x01, 0x02, 0x03, 0x04, 0x05, 0x08, 0x07,
    0x08, 0x09, 0x0a, 0x0b, 0x0c, 0x0d, 0x0e, 0x0f],
```

so that nobody gets the impression that this is a "randomly-generated" key.

crypto/aes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
...
+//!     KeyType::SymmetricCipherKey,
+//! ).expect("a 16-byte symmetric cipher key");
+//!
+//! let aes = AES_128::new(&key).expect("a valid AES-128 key");
+//!
+//! // FIPS 197 Appendix B.
+//! let mut block = [0x32, 0x43, 0xf6, 0xa8, 0x88, 0x5a, 0x30, 0x8d,
+//!                  0x31, 0x31, 0x98, 0xa2, 0xe0, 0x37, 0x07, 0x34];
```
ounsworth: Since this is supposed to be illustrative, maybe add a comment explaining what that plaintext is. Is it an ascii string for something? If not, then maybe just

```
// Some arbitrary data
```

crypto/aes/src/lib.rs
```diff
@@ -0,0 +1,237 @@
+//! A constant-time, table-free AES block cipher engine (NIST FIPS 197).
+//!
+//! This crate provides the raw AES keyed permutation -- [`AES_128`], [`AES_192`] and [`AES_256`] --
+//! implemented as a Boolean circuit over bit-planes rather than as byte substitutions through a
+//! lookup table. That makes it both smaller and constant-time; see [Design](#design).
+//!
+//! It is a *permutation*, not a cipher you can encrypt data with. See
+//! [Security Considerations](#security-considerations).
```
ounsworth, line 8: But this crate now does contain modes, and in fact they are described below in this very doc block! So I think this whole docs block needs an editing pass for consistency.

crypto/aes/src/sbox.rs
```diff
@@ -0,0 +1,381 @@
+//! SUBBYTES() and INVSUBBYTES() as a Boolean circuit (FIPS 197 Sec 5.1.1 and Sec 5.3.2).
```
ounsworth, line 1: Question: these are module docs "//!", but lib.rs declares this as a private module, so they will never build. So what's the point?

Also, these docs duplicate some of what's already been said in the lib.rs docs and could be slimmed down quite a bit.

crypto/aes/src/bitslice.rs
```diff
@@ -0,0 +1,210 @@
...
+//!
+//! The three-stage masked-swap transpose and the even/odd two-block packing are translated from
+//! BearSSL `src/symcipher/aes_ct.c` (`br_aes_ct_ortho`) and `aes_ct_cbcdec.c` (the `q[0]`,
+//! `q[2]`, `q[4]`, `q[6]` load order), by Thomas Pornin, MIT licensed.
+
+/// One 16-byte AES block, in the order of FIPS 197 Eq (3.6): `block[r + 4c] == s[r,c]`.
+pub type Block = [u8; crate::BLOCK_LEN];
+
+/// The eight bit-planes holding two blocks. See the module docs for the layout.
+pub(crate) type Planes = [u32; 8];
```
ounsworth, line 63: You are defining the SBox circuit to act on 2 blocks (32 bits) at a time.
Given that almost all CPUs these days are 64-bit, couldn't you do 4 blocks at a time (ie `Planes = [u64; 8];` for the same cost, twice the throughput, and without changing any of the code in sbox.rs?

I actually wonder if we could be a bit clever here and make `Planes<T>: [T; 8];` where `T` can be either u16 (used by `encrypt()`), u32 (used by `encrypt_2blocks()`), or u64 (used by `encrypt_4blocks()`)? That way, you'd get maximum efficiency on the 4blocks on a 64-bit machine, but on a 32 or 16 bit machine you don't lose anything in the single-block mode.

Also, only some modes have parallelizable blocks that can even take advantage of this, right? In particular CBC and GCM cannot, right? So a `Planes<T>;` means that you can always process it at exactly the size you need without needing to run it half empty, which is at least wasteful on memory usage, if not also affecting performance.

dghgit: Yes.There's a four block version... I've wondered about that, there's also a lot of 32 bit stuff out there though, so I thought I'd run with Thomas Pornin's insites on this one. The other 4 are all 64 bit at the moment (I'd be tempted to try and get TDES back to 32 as well...). This isn't a change I'd make now though, there are multiple answers to this question (maybe offer 32 and 64, maybe move to 64, maybe stay on 32) we really need to try and collect some data.

crypto/aes/src/lib.rs
dghgit: No argument from me, feel free.

crypto/aes/src/sbox.rs
dghgit: So, I think this stuff is a mixture of LLM's writing notes for themselves and also the QUALITY_AND_STYLE.md document. In this case I think I'd leave them in place. Whoever/whatever looks at the file later might find them helpful, whether they build or not probably doesn't matter. I think we do need to keep in mind that these days we're writing for two audiences (not that I'd ever put myself forward as a great example of pro-active commenting/documentation, being famous for extensive use of the phrase "The code is the documentation"...)

crypto/aes/src/lib.rs
dghgit: Sure, again, feel free.

crypto/core/src/traits.rs
dghgit: bouncycastle-cipher works for me. We could have bouncycastle-pqc, bouncycastle-pk as well I guess... that would bring the top level count down.

crypto/core/src/traits.rs
dghgit: I've done this, I've also  made one other change in connection with this, when I added the Ecb wrapper I actually ignored the RNG (yes in a sense ECB is our very own Rot 13...) I checked bc-java though, we've always traditionally thrown an exception, so I think the panic() behaviour is correct - passing ECB an RNG is not going to save you... and I'm pretty sure anyone doing it will have hit tab at the wrong time and instead of getting _CBC will have ended up with _ECB. No good will come of that!

crypto/core/src/traits.rs
ounsworth: We could also have a single bouncycastle-crypto.

This question about packaging is good to ask at this stage in the project before it gets locked in. And maybe I'll take it over to the bc-rust channel in matrix to get everyone's opinion.

My thinking has been roughly that it's worth creating a new crate when:
1. This is something that someone would want to use in isolation from the rest of the library as a direct dependency? Yes for complete standalone algorithms like Hex, SHA2, AES, MLKEM, etc. No for helpers / building blocks like KeyMaterial, ConstantTime, HMAC<HASH> -- although, maybe debatable for HMAC<HASH>, because someone could impl their own HASH and then use our HMAC<HASH> on top, but that's probably rare.
2. Is this building blocks common to multiple other crates that doesn't really fit with any of the other building block crates?

To me, cipher padding and cipher modes are not standalone algorithms that someone would want to use independently from the rest of bc-rust, and they could logically go together into some kind of cipher building blocks crate as `bouncycastle-cipher::modes::*` and  `bouncycastle-cipher::padding::*` ... I can't think of a reason **not** to do that, like why packaging them together would be confusing.



(Aside: "bouncycastle-pqc" ... at what point do we stop calling that "pqc" and we just call it "crypto"? I would argue that we're there ...)

crypto/core/src/traits.rs
ounsworth: Now I'm thoroughly confused.

On this branch, there exist definitions for `SymmetricCipherEncryptor::encrypt()`, `SymmetricCipherEncryptor::encrypt_ou()`, etc.

But you're saying:

> We could write a "Cipher::encrypt( b"Attack at dawn!" )" but you wouldn't do it here, it would be a general purpose utility built on SymmetricCipherEncryptor/Decryptor and AEADCipherEncryptor/Decryptor

That doesn't make any sense to me because it is being done here already.

Remember that you can't (elegantly) do function overloading in rust. So if this trait is defining `::encrypt()`, then this trait is where `::encrypt()` is defined. Making more traits that also define a `::encrypt()` is going to be confusing and hard to use. I don't want function overloading across multiple traits; that ends up being absolutely hideous to program with.

I also don't understand why this trait `SymmetricCipherEncryptor` can't **be** the umbrella trait that applies to block ciphers, stream ciphers, and aeads into a singly API? Maybe we need a phone call to sort this out.

dghgit: So the issue, in as much as there is one, is AEAD ciphers can use associated data and the tags are also expressed as detached, rather than always inline.

That said, if you wanted to offer SymmetricCipherEncryptor as supporting a **subset** of an AEAD cipher, I think we could get that to work, we'd just need to do something like:

```
pub trait AEADCipherEncryptor<....TAG_LEN>: SymmetricCipherEncryptor<....> {

     fn do_update_aad(&mut self, aad: &[u8]) -> Result<(), SymmetricCipherError>;

    fn do_final_out_detached(...) -> Result<([u8; NONCE_LEN], usize, [u8; TAG_LEN]), SymmetricCipherError> { ...}
     
    // other stuff...
}
```

then (in theory) SymmetricCipherEncryptor::encrypt()  would produce cipher text from an AEAD instance which simply have no associated data and an inlined tag, and the AEAD trait would have a encrypt_detached() method for the case where there was no associated data and a detached tag.

Is that the effect you're looking for?

We could make the preliminary changes to the AEAD PR to support this now, but doing it will mean waiting till the AEAD PR is merged as well, so I'd treat it as out of scope for this PR right now.

Not quite an umbrella, more like holding a magazine over your head, so only partial coverage.

crypto/aes/src/lib.rs
```diff
@@ -0,0 +1,239 @@
...
+mod cfb;
+mod cfb8;
+mod ctr;
+mod ecb;
+mod padded_mode;
+mod round;
+mod sbox;
+mod schedule;
+
+pub use aes::{AES_128, AES_192, AES_256, BLOCK_LEN};
```
ounsworth: I propose that these actually shouldn't need to be public. They are building blocks, not something that we want an amateur finding in their pop-up search window and using. So let's hide them.

dghgit: Sure, anything we can do to lock things down seems good to me.

crypto/aes/Cargo.toml
```diff
@@ -0,0 +1,23 @@
...
+version.workspace = true
+edition.workspace = true
+
+[dependencies]
+bouncycastle-core.workspace = true
+bouncycastle-utils.workspace = true
+# Only for the AES-CBC type aliases in `cbc.rs`; the engine itself does not use it.
+bouncycastle-modes.workspace = true
+# Only for the padded AES-CBC aliases in `cbc.rs`; the engine itself does not use it.
+bouncycastle-padding.workspace = true
```
ounsworth: Note-to-self: remove the comments, then try a branch where these are merged into a single `bouncycastle-cipher` crate.

PR #135, opened 2026-09-16
Ensure that core and core-test-framework crates support no_std

jjkurczak:
## Issue Link
resolves bcgit/bc-rust#134

## Summary
Updates core and core-test-framework crates to build with no_std, and configures workspace and other packages to support no_std

### What did you do?

This PR incrementally enables no_std support for bouncycastle-core and bouncycastle-core-test-framework. The most consequential changes are:

1. Adding a new `alloc` feature to a number of crates in the dependency graph of bouncycastle-core in tests mode (hkdf, hmac, rng, sha2)
1. Adding `#![cfg_attr(not(feature = "alloc"), no_std)]` to `crypto/core/src/lib.rs` to force no_std when alloc is not enabled for bouncycastle-core
1. Updating `crypto/core/src/traits.rs` to feature-gate any std/alloc dependent functions in the core bc-rust traits
1. Adding feature gates to various impls that depend on `traits.rs` in the crate dependency graph
1. Adding `_array()` functions in `traits.rs` for convenient alternatives to `_out()` functions when in no_std mode, with convenient default implementations that function as a wrapper over the existing `_out()` versions.
1. Updating bouncycastle-core-test-framework to feature-gate test types that are not applicable for no_std (feature-gated trait functions)

### What alternatives did you consider?

1. I had initially attempted to create a large PR that would cover a few crates at once, but this was too large and accruing merge conflicts
1. I've chosen to create even smaller, incremental PRs that enable no_std (and no alloc) one or two crates at a time to keep reviews manageable, because the scope of the changes to manage no_std is quite large when alloc is also disallowed
1. Although we also intend to support a no_std WITH alloc mode, I think it makes more sense to fully remove std and alloc first since this has bigger implications for the architecture, and add alloc back in later as a smaller step
1. A number of tests are feature-gated in core-test-framework because they are only testing trait functions available in std. Many of these tests do not have duplicates for no_std equivalent behaviours, and these have been noted with `// todo: may require no_std equivalent` in the code. I believe it will be more efficient to manage adding back parallel tests in their own PR because it will take me some effort to understand and replicate
all of them correctly.
1. All tests-only code (i.e., integration tests in tests/, bouncycastle-core-test-framework) is not required to support no_std. This would be a lot of trouble to manage, and there is no need to run the tests themselves in no_std environments.

### Did you add or update any tests to account for these changes?
No tests added - many tests feature-gated for no_std builds, see reasoning above.

### Is there anything else the reviewer should keep in mind?
1. For the moment, only core and core-test-framework can build, test, or benchmark individually with --no-default-features. Building with --no-default-features, or otherwise disabling `alloc`, for any package other than sha2 or sha3 (including when building --workspace), is currently an *unsupported configuration*.
1. No changes have been made to github actions to validate --no-default-features builds - it seems like this should be managed once the entire workspace of packages is capable of being validated for --no-default-features? Is it worth spending the effort to mess around with the github actions for each sub-issue, so each set of packages is validated as they begin to support no_std, and then refactor for the entire workspace functioning at the end?
1. bouncycastle-core-test-framework does NOT enforce no_std in its crate, even without the "alloc" feature, because it will only run for tests/dev-builds
1. Some new default implementations in traits.rs (Hash::do_final_partial_bits_array, XOF::hash_xof_array, and XOF::squeeze_array) do not have any tests defined yet. Although these are all only wrappers for the existing `_out()` equivalents, these should be covered when the parallel implementations are added for the tests that have been feature-gated without any no_std equivalents.

### Which packages are impacted?
- This change impacts core, core-test-framework, hkdf, hmac, rng, sha2.
- Nearly all changes relate to disabling code when the `alloc` feature is not enabled, so default builds using `alloc` should ideally see no change.

### What runtime behaviors could be impacted?
 It's unlikely that any changes in this PR will impact runtime behaviours - I would expect to see breaks at compile time rather than at runtime.

### What could go wrong in the worst-possible scenario?
There are two major areas of risk I've identified:
1. There are widespread changes to workspace and package .toml files to create the `alloc` feature in many packages, and to ensure that dependencies only inherit `alloc` when needed. There could be unexpected dependency breakages when bulding packages if the dependencies have been misconfigured.

## Validation
1. build, test, bench workspace should continue to function as expected (bench manually)
1. build, test, bench individual packages should continue to work as expected (build, test, bench manually)
1. build, test, bench core and core-test-framework packages individually with --no-default-features flag should work, with reduced test coverage (test manually)

## AI Usage Statement
- [ ] No
- [X] Yes, non-trivial code changes were generated by AI

Assisted-by: ClaudeCode:Sonnet5


ounsworth:
Comments on the PR description:

> For the moment, only core and core-test-framework can build, test, or benchmark individually with --no-default-features. 

Are you proposing to merge this to the release branch like this, or build a stack of PRs on top of this and merge when everything is complete?

---

> Some new default implementations in traits.rs (Hash::do_final_partial_bits_array, XOF::hash_xof_array, and XOF::squeeze_array) do not have any tests defined yet.

Implementations for those are coming in on David's #133 .

---

> There are widespread changes to workspace and package .toml files to create the alloc feature in many packages, and to ensure that dependencies only inherit alloc when needed. 

This may be softof a note-to-self, but we should consider (and document) how this behaves both in the context of a build within our workspace, and where another workspace has a dependency on one of our crates (ie the way this will eventually be consumed on crates.io).

crypto/base64/Cargo.toml
```diff
@@ -4,12 +4,12 @@ version.workspace = true
 edition.workspace = true
 
 [dependencies]
-bouncycastle-core.workspace = true
+bouncycastle-core = { workspace = true, default-features = true }
```
ounsworth, line 7: You've done more reading / playing into cargo features than me, but if you have to write "default-features=true" everywhere, doesn't that defeat the purpose of them being defaults?

The way I read this, if you specify nothing, then it should get the defaults?
https://doc.rust-lang.org/cargo/reference/features.html#the-default-feature

crypto/core-test-framework/src/hash.rs
```diff
@@ -28,44 +32,68 @@ impl TestFrameworkHash {
...
 
-        /*** fn hash(self, data: &[u8]) -> Vec<u8> **/
-        let output_vec = H::default().hash(input);
-        assert_eq!(output_vec, expected_output);
+        #[cfg(feature = "alloc")]
+        {
+            /*** fn hash(self, data: &[u8]) -> Vec<u8> **/
+            let output_vec = H::default().hash(input);
+            assert_eq!(output_vec, expected_output);
+        }
```
ounsworth, line 40: That pattern makes sense. I could see this getting messy if we have lots of overlapping features, but checks out 👍

crypto/core/src/traits.rs
```diff
@@ -16,7 +21,7 @@ use crate::key_material::KeyType;
 pub trait AEADCipher<const KEY_LEN: usize, const NONCE_LEN: usize, const TAG_LEN: usize>:
     SymmetricCipher<KEY_LEN, NONCE_LEN> + Sized
 {
-    #[cfg(feature = "std")]
+    #[cfg(feature = "alloc")]
```
ounsworth, line 24: This is a really nit-picky style thing, but is it more idiomatic to do:

```rust
#[cfg(feature)]
/// Docstring comments
fn do_something()
```

or


```rust
/// Docstring comments
#[cfg(feature)]
fn do_something()
```

We're probably a bit inconsistent about this across the library at the moment.

I think I have a weak preference for the latter, especially since docstring comments can sometimes go on for hundreds of lines. We could probably ask Claude if there's an overarching convention for this in the rust community.

crypto/hmac/Cargo.toml
```diff
@@ -3,18 +3,27 @@ name = "bouncycastle-hmac"
...
 edition.workspace = true
 
+[features]
+default = ["alloc"]
+alloc = [
+    "bouncycastle-core/alloc",
+    "bouncycastle-core-test-framework/alloc",
+    "bouncycastle-rng/alloc",
+    "bouncycastle-sha2/alloc",
+]
```
ounsworth, line 13: I like this too -- declaring only alloc and not all of std by default means that it won't accidentally drag the whole std static lib into a calling application 👍

ounsworth (changes requested):
I like this pattern -- we're setting up a framework where crates are minimally functional in a no_std, no alloc build, and add more user-friendliness features in an alloc or a full std build.

Some general comments:

---

This is setting project architecture for how we're handling no_std and alloc. I suggest you add a section to QUALITY_AND_STYLE.md > Architecture laying out the design pattern for future contributors (human and AI).

---

Since this is something of a selling feature for the library that will distinguish us from other libraries, might also be worth putting something in the top-level README.md.

---

I am a big fan of crate docs; I am a big believer that the public docs are an integral part of the library's API surface.

One thing missing from this PR is how we are communicating the std / alloc / no_std differences to users via the crate docs. QUALITY_AND_STYLE.md already lists a few mandatory docs sections that should appear in each crate:

* Usage Examples
* Memory Usage
* Security Considerations

I wonder if each crate should also carry a "# std and alloc" that could just say "This crate requires std", or provides a brief description of the std and alloc cargo features that this crate carries, what functionalities will be disabled by removing the default features and how to actually perform a command-line or cargo build with those features off. This will likely end up being a lot of standard boilerplate across crates, but I think it's probably worth duplicating it across crates so that each crate's docs are self-contained.

Also, we probably need to establish a docs convention for tagging functions that will disappear in an alloc or no_std build.

---

I know that @npajkovsky has experience here, so it's important to me that he's on-board with this approach.

crypto/base64/Cargo.toml
jjkurczak: If you specify nothing, what is enabled will depend on every other crate that also depends on bouncycastle-core. The bouncycastle-core crate will be built exactly once for any cargo build, with the set of enabled features being the union of all feature requests from all dependent crates.

I believe this particular change WAS actually necessary for a particular crate build to not break, but I can't remember which. I'll double-check on this one.

crypto/core-test-framework/src/hash.rs
jjkurczak: Yeah, it's more or less going to be the cleanest way to do it I think!

crypto/core/src/traits.rs
jjkurczak: Yes, I struggled with this a little, and I don't know. I think it looks cleaner to have the cfg at the top, and that's also what I'm seeing here:

(https://github.com/QUBIP/openssl-provider-forge-rs/blob/7fb96bb531d6120586da097fcf0168c82aa847d4/src/upcalls.rs#L41)

But if the docstring comments are very long that does make it hard to connect. I'm also wondering if docstring comments that go on for hundreds of lines is actually an anti-pattern anyway?

crypto/hmac/Cargo.toml
jjkurczak: Sorry, I just want to check whether this is unclear. Right now, there is one feature being added called "alloc" that currently gates both "std" and "alloc" in this PR. This PR is essentially enabling the "innermost circle" of no_std, and no alloc, for the sake of simplicity and incremental change.

It will be future work to separate the std gates and alloc gates, and add an "std" feature to be able to disable std without disabling alloc.

jjkurczak:
@ounsworth 
> Comments on the PR description:
> 
> > For the moment, only core and core-test-framework can build, test, or benchmark individually with --no-default-features.
> 
> Are you proposing to merge this to the release branch like this, or build a stack of PRs on top of this and merge when everything is complete?

I am proposing to merge each incremental change into the release branch. This is to:
1. ensure that individual PRs are self-contained, complete, and correct
1. be able to show incremental progress of completed work towards the goal
1. prevent the kind of horrible merge conflicts that a single no_std branch would accumulate (I am estimating that completing all of this work is on 4-6 week timeline).

> > Some new default implementations in traits.rs (Hash::do_final_partial_bits_array, XOF::hash_xof_array, and XOF::squeeze_array) do not have any tests defined yet.
> 
> Implementations for those are coming in on David's #133 .

Do you mean for the _out() versions of these? Because _array() versions I think are only being newly added in this PR?

> > There are widespread changes to workspace and package .toml files to create the alloc feature in many packages, and to ensure that dependencies only inherit alloc when needed.
> 
> This may be softof a note-to-self, but we should consider (and document) how this behaves both in the context of a build within our workspace, and where another workspace has a dependency on one of our crates (ie the way this will eventually be consumed on crates.io).

Distributing individual crates is an unsolved problem for this repo that I don't yet know how to address. Currently all of the crates inherit dependencies from the workspace, but that doesn't work for downloading an individual crate. I think this is another issue that needs to be addressed iteratively, and I'm not sure if it belongs in a sub-issue for the no_std issue, or in another github issue altogether.

crypto/base64/Cargo.toml
jjkurczak: Sorry, yes I'm remembering now, it was slightly more complicated than this. What I said above is correct, HOWEVER, there's another wrinkle that was affecting crates when built individually.

Adding `default-features = true` ends up being a special interim requirement for any crate that doesn't yet have its own `alloc` feature, but REQUIRES its dependencies to be built with `alloc` enabled. Since the workspace itself sets `default-features = false` for all crates that DO support the `alloc` feature, every other crate inherits those dependencies with `default-features = false`. However, crates that have their own `alloc` feature are configured to pass their feature setting on to their dependencies so it's not a problem for them.

And unfortunately it appears that we need to use `default-features = false` in the workspace dependencies, because otherwise trying to set `default-features = false` in individual crates conflicts somehow, and then cargo complains and quits. I'm a little vague on that part, I'd be happy to find another way to do it! But also, this also reinforces that we need to figure out how this can all work without a parent workspace when consuming individual crates.

npajkovsky:
So, I have something for you, but you may not like it. The strongest proof that bc-rust works with `no_std` is to build it for a target that has **no** std at all. Such a target is `thumbv7em-none-eabi`, following commands will give you tons of errors, and I would start with submitting PRs for `std::result::Result -> core::result::Result;` and so on.

```
rustup target add thumbv7em-none-eabi

cargo build \
  --workspace \
  --exclude cli \
  --exclude mem_usage_benches \
  --exclude bouncycastle-core-test-framework \
  --lib \
  --no-default-features \
  --target thumbv7em-none-eabi
```

This selects the production libraries, including the umbrella crate, while excluding the host-side CLI, benchmarking package, and test framework, which requires `std`.

I suggest adding this target to CI so it's continuously tested. 

We also need consistent feature definitions and forwarding across the crates. For example, schematically:

Root Cargo.toml:
```toml
[features]
default = ["std"]
std = ["alloc", "bouncycastle-sha2/std"]
alloc = ["bouncycastle-sha2/alloc"]

[dependencies]
bouncycastle-sha2 = { workspace = true, default-features = false }
```
crypto/sha2/Cargo.toml:
```toml
[features]
default = ["std"]
std = ["alloc"]
alloc = []
```

The same feature forwarding must extend through each crate’s dependencies where applicable. Each library would use:
```rust
#![cfg_attr(not(feature = "std"), no_std)]

#[cfg(feature = "alloc")]
extern crate alloc;
```

Consumers could then select the minimal configuration:
```
bouncycastle = { version = "...", default-features = false }
```
Or enable allocation while keeping std disabled:
```
bouncycastle = { version = "...", default-features = false, features = ["alloc"] }
```
This requires that no other dependency enables those crates' std features, because Cargo unifies feature requests.

We may also consider writing a small bare-metal `#![no_std]` executable that calls representative library APIs. It supplies the startup/panic handling and, for allocation-enabled configurations, an allocator. Library builds establish compilation; the executable additionally checks final-link requirements. 

Or maybe even better, we can create an interop test with rustle that implements all features under no_std too. This step needs more thinking.

There should also be some thoughts on why we have two of the same API.
```rust
#[cfg(feature = "alloc")]
fn hash(self, data: &[u8]) -> Vec<u8>;
```
and
```rust
#[cfg(feature = "alloc")]
fn hash(self, data: &[u8]) -> Vec<u8> {
    let mut output = vec![0u8; PARAMS::OUTPUT_LEN]; // self.output_len()
    self.hash_out(data, &mut output);
    output
}
```
`hash()` and `hash_out()` does the same thing, however `hash_out()` works with no_std. If we want to have `hash()` with alloc, then, if possible, it should be a default trait method.

PR #136, opened 2026-09-17
Mikeo/lowmem improvements

ounsworth:
In short, I had Claude do a review for silent copies in the mlkem and mldsa lowmemory crates and it found a couple of instance of:

* moves that were being compiled as a pass-by-value copy
* .try_into() that created a copy as a result of needing an intermediate owned copy.
And similar patterns.
* Adjusted the size of the SHAKE pre-buffer in SampleNTT() / RejNTTPoly() to match exactly a SHAKE block size boundary.

The net result is about a 2 - 4 kb reduction across the whole lowmemory suite, and also some tightening up in the non-optimized implementation.


ounsworth:
Thanks @dghgit for the review. I've updated to address. Ready for re-review.

npajkovsky (approved):

PR #137, opened 2026-09-18
Security issue in ML-DSA: missing reduce32 leads to i32 overflows 

ounsworth:
https://eprint.iacr.org/2026/1032, Wycheproof mldsa_87_verify tcId 240/241

Performance impact of this fix is not measurable by our unit tests: two bitshifts and a couple additions is negligible compared to the InvNTT that's about to follow.

Assisted by: Claude Fable 5.1


dghgit (approved):
Sorry, I thought I did this one a couple of days ago.

The code's fine but it looks like the project guidelines and the house style have drifted, and CLAUDE.md needs some work to deal with it, since you're running house style, I leave it with you, it understandable this would be changing, but we need to keep the supporting docs in sync otherwise both people and tools will get confused - once the "paperwork issues" are fixed in the PR, this should be fine to merge. The report pr137_review_3123379.md details what's getting the tools into trouble. There is no need to request a further review once the report is dealt with.

After that, given what it fixes it should go straight onto 0.1.3alpha, I can run a job to then merge it upstream.

PR #144, opened 2026-09-20
doc: clarify how to build bc-rust code

mcr:
-b cli is needed to build bc-rust executable.


ounsworth:
Right!

That's because the top-level [cargo.toml](https://github.com/bcgit/bc-rust/blob/main/Cargo.toml) includes `cli/` in the workspace, but it's not listed as a dependency of the library (because in fact the dependency is the other way around: cli depends on the library).

Yup, change == good.

ounsworth (approved):
Marked Approved, but we'll wait to merge until we've de-log-jammed the cluster of PRs.

npajkovsky (approved):

PR #143, opened 2026-09-19
ec, ecdsa, sm2: custom-curve arithmetic, ECDSA and SM2 signatures

dghgit:
## Summary
Adds `bouncycastle-ec` (constant-time custom-curve field/point/scalar arithmetic for P-256/384/521, secp256k1, Brainpool 256/384/512, and SM2), `bouncycastle-ecdsa` (ECDSA over all of those curves, RFC 6979 deterministic nonces), and `bouncycastle-sm2` (the SM2 signature scheme). This is a port of Bouncy Castle Java's custom-curve (`org.bouncycastle.math.ec.custom`), Brainpool, and SM2 implementations.

## Description
- Each curve has its own fixed-width field/scalar types (no `BigInteger`-style variable-width arithmetic on secret data), Jacobian point arithmetic, and a fixed-base comb for `[k]G` plus wNAF for variable-base multiplication.
- ECDSA: RFC 6979 deterministic `k`, DER and raw `r || s` encodings, CAVP + Wycheproof vectors per curve.
- SM2: GB/T 32918.2-2016 signing/verification with the standard's `ZA` user-identity hash.

## Scope and Risk
New crates only, layered on top of `feature/xof-cshake`.

## Validation
- `cargo test --workspace`: all passing (CAVP P-256/384/521, Wycheproof per curve, RFC 6979 vectors, SM2 KAT vectors).
- Mutation testing was already run per curve/file during development (e.g. P-521 field/point/scalar: 336 mutants, 250 caught, 55 unviable, 21 missed; secp256k1: 270 mutants, 195 caught, 56 unviable, 14 missed; each brainpool curve and its ECDSA wiring similarly), with findings fixed along the way -- remaining misses are the XOR/OR-equivalence class this repo accepts. `cargo mutants -p bouncycastle-sm2` (the whole crate): 98 mutants, 52 caught, 46 unviable, 0 missed.
- `cargo mutants -p bouncycastle-ecdsa` (full crate, completed): 921 mutants, 538 caught, 371 unviable, 3 missed, 9 timeouts, 0 failures. The 3 missed are XOR/OR equivalences; the 9 timeouts are all `hmac_k -> Secret::new()` stub mutations across the seven curves' RFC 6979 files, which create a genuine non-terminating reseed loop rather than a fast failure -- effectively kills. See the PR comment for the breakdown.
[...]

PR #146, opened 2026-09-20
rsa: RSASSA-PKCS1-v1_5, RSASSA-PSS and PSS-SHAKE across RSA-1024 through 8192

dghgit:
## Summary
RSA digital signatures only -- no encryption or KEM yet -- across a limited range of modulus sizes up to the maximum originally requested, 8192 bits: RSASSA-PKCS1-v1_5, RSASSA-PSS (RFC 8017 §8), and PSS-SHAKE128/256 (RFC 8702 §3.2.1) over RSA-1024/1536 (verify-only) through RSA-2048/3072/4096/8192 (sign+verify), exposed through `bouncycastle_core`'s `Signer`/`SignatureVerifier`/`SignaturePrivateKey`/`SignaturePublicKey` traits in the same shape as the ECDSA, SM2 and ML-DSA crates. Key pair generation (FIPS 186-5 Appendix A.1.3) is included for the signing sizes. Hopefully a starting point a later PR can extend to RSAES-OAEP/PKCS1-v1_5 encryption and an RSA-KEM.

## Description
- Reuses `bouncycastle_ec`'s constant-time fixed-width limb arithmetic (`bouncycastle_ec::nat`, `bouncycastle_ec::montgomery`) rather than a dedicated BigInteger type, for the same constant-time-on-secret-exponent reasoning that motivates that crate's own custom-curve arithmetic: a modular exponentiation over a secret exponent has to run in the same time regardless of the exponent's value, which needs limbs of a fixed width known at compile time.
- 2048-8192 bit moduli support both signing and verification; 1024- and 1536-bit moduli are verification-only, enforced by the absence of a private-key type for those sizes rather than a runtime check -- and, for the traits, by `Signer` only being implemented where the private key type implements `SignaturePrivateKey` (a `compile_fail` doctest in `rsa_1024.rs` pins it).
- API shape (commits `f1a87e2` .. `02f0060`, in review order): the EMSA/RSASSA layers take the message hash rather than the message (own refactor commit), so the streaming trait impls share one encoding path; one generic streaming type per scheme (`RSASSA_PKCS1_v1_5`, `RSASSA_PSS`, `RSASSA_PSS_SHAKE`, spec spelling per QUALITY_AND_STYLE) with per-size aliases (`rsa_2048::RSASSA_PKCS1_v1_5_SHA256`, ...); key traits implemented per size at that size's `SK_LEN`/`PK_LEN`; `ctx` accepted-but-ignored as ECDSA does. PSS randomness follows the other crates: `Signer::sign` draws the salt from the default RNG (the trait's own contract), `sign_randomized(sk, msg, rng)` takes a caller RNG (ECDSA/SM2's shape), and `set_signer_salt` on a streaming state fixes it (ML-DSA's `set_signer_rnd` shape). `SignatureVerifier::verify`'s `&[u8]` signature makes RFC 8017 §8.1.2/§8.2.2 step 1's length check a runtime check, and step 2.b's RSAVP1 range failure is folded into "invalid signature" on the trait path (the generic `*_from_hash` functions keep RSAVP1's own error for vector-driven callers).
- The per-size free functions (`pkcs1_v1_5_sign_sha256` and ~70 siblings) were removed once the traits covered them (commit `05988a2`), and the key type aliases renamed to `RSA2048PrivateKey`/`RSA2048PublicKey` (etc.) to match `ECDSAP256PrivateKey`/`MLDSAPublicKey`/`SM2PrivateKey` (commit `02f0060`).
- Key pair generation (commit `1efdd64`): FIPS 186-5 Appendix A.1.3 (probable primes: trial division against the primes below 2^10, then Appendix B.3.1 Miller-Rabin with Table B.1's round counts; `e = 65537`; `|p - q|` and `d > 2^(nlen/2)` criteria from A.1.1), structured after BC Java's `RSAKeyPairGenerator`, on the crate's own limb arithmetic -- no bignum dependency. Exposed as `rsa_<size>::keygen()`/`keygen_from_rng()` (the ECDSA shape, and what `core-test-framework` takes) plus a public `keygen::is_probable_prime`, and as `rsa-<size> keygen --pkfile` in the CLI. Generation is documented as not constant-time (rejection sampling); signing is unaffected.
- CRT key types and RSASP1/RSAVP1 (RFC 8017 §5.2), criterion benches (including keygen at 2048/3072), a stack-usage harness, and CLI subcommands, which stream stdin through the traits in this CLI's usual 1 KB chunks.

## Scope and Risk
New crate only (`bouncycastle-rsa`) plus its CLI wiring, layered on top of `feature/ecdsa` (uses `bouncycastle_ec::nat`/`montgomery`, and carries `feature/simple-ciphers`'s cipher-return-type/`BlockCipherPadding` rename transitively via that merge). Nothing outside the new crate and its CLI files changes.

## Validation
- `cargo test --workspace`: all passing. For the crate: Wycheproof PKCS1-v1.5/PSS vectors per size/hash (the RSA-2048/SHA-256 verify file re-run through `SignatureVerifier::verify` so its wrong-length vectors reach step 1's check), BC Java cross-check vectors at 8192 (also through the trait path), `core-test-framework`'s `Signer`/`SignatureVerifier` conformance suite for every pairing at 2048/3072 (with the exhaustive bit-flip pass once per generic type) and for the SHA-256 pairings at 4096/8192 with round trips for the rest, `test_keys` at every signing size, `set_signer_salt`/`sign_randomized` against the generic `sign_from_hash_with_salt` and a fixed-output RNG; for key generation, Miller-Rabin known answers (Wycheproof's `p`/`q`, M521, M607 accepted; `p+2`, `p+4`, `p+6`, `2^521+1` rejected), FIPS 186-5 A.1.1's criteria checked on generated RSA-2048/3072 pairs (and RSA-4096/8192 under `--ignored --release`, about 30 s together) plus sign/verify round trips, the conformance suite with the real generator, and the weak-RNG and stuck-RNG failure modes; `cli/tests/rsa_cli_tests.rs` drives the binary (RSA-1024 Wycheproof verification, RSA-2048 PKCS#1 v1.5 byte-for-byte against the library over a multi-buffer payload, fresh PSS/PSS-SHAKE signatures verified both ways, tampered/truncated signatures and wrong-length key files rejected).
- `./dev_scripts/quality_stats.sh ./crypto/rsa`, before/after the trait work: core code 2270 -> 2393 lines, docstring lines 766 -> 858, runtime fallibility unchanged (1 unwrap, 20 `Err()`; the removed doc example accounts for the other unwrap). Key generation adds 435 core lines and 4 `Err()` sites (RNG strength, RNG failure, the two candidate limits); its Miller-Rabin round count is a `NonZeroUsize` rather than a runtime check.
- `cargo mutants -p bouncycastle-rsa` on the pre-trait crate (through `61aa840`), done in two passes so the numbers here are an honest account rather than a single clean run:
  - An initial full-crate run (648 mutants) was intentionally stopped at 499/648 (363 caught, 20 missed, 3 timeout, 113 unviable) once every finding it had reached was investigated and either fixed (new tests, commit e30536d) or documented as an equivalent mutation.
  - A scoped follow-up re-tested exactly the 7 files touched by those fixes, to completion: 353 mutants, 229 caught, 6 missed, 4 timeout, 114 unviable. The 6 missed are documented OR/XOR and boundary-case equivalences; the 4 timeouts are one genuine infinite loop plus 3 confirmed (empirically, in an isolated worktree) to be harness artifacts -- `cargo test` doesn't stop at the first failure, so a fast-failing mutant still waits on the rest of a heavy, forced-single-threaded suite before `cargo test` returns nonzero.
  - A full-crate confirmation run of that pre-trait state is in progress on separate hardware, mirroring PR #143 (`bouncycastle-ec`); its result will follow as a PR comment.
- The trait-wiring and key-generation commits (`f1a87e2`..`1efdd64`) have **not** been mutation-tested yet; a scoped `cargo mutants` run over the 14 files they touch is in progress and will be reported as a PR comment.

## AI Usage Statement

Assisted-by: Claude:claude-sonnet-5
Assisted-by: Claude:claude-fable-5-1

🤖 Generated with [Claude Code](https://claude.com/claude-code)


dghgit:
## cargo mutants -- full-crate confirmation run, pre-trait-wiring state (`61aa840`)

Completed on separate hardware (mirroring PR #143's approach): **646/646 mutants**, 446 caught, 6 missed, 3 timeout, 191 unviable.

All 9 non-caught findings are already-documented equivalents/artifacts, consistent with the two-pass investigation described in the PR body:

**Missed (6), all documented OR/XOR or boundary equivalents:**
- `emsa_pkcs1_v1_5.rs:124:18`, `:151:28`, `:192:20` (`decode_digest_info`/`emsa_pkcs1_v1_5_verify`) -- boundary-condition and short-circuit equivalents already covered by the crate's own vector-driven tests.
- `modexp.rs:109:54`, `:185:36`, `:193:54` (`redc_normalize`/`ct_shift_in_bit_mod`) -- disjoint-bit-range OR/XOR equivalences (the two operands never overlap by construction, so `|` and `^` agree).

**Timeout (3), genuine infinite loops, not coverage gaps:**
- `emsa_pkcs1_v1_5.rs:182:11` (`emsa_pkcs1_v1_5_verify`) -- a `+=`→`*=` mutation stalls a length-tracking loop forever.
- `mgf1.rs:24:18`, `:34:16` -- similarly stall MGF1's output-generation loop.

This confirms the pre-trait-wiring baseline (everything through commit `61aa840`) is clean: no silent gaps, every non-caught mutant independently verified via code reasoning to be either a strict semantic equivalent or a loud (non-hanging-forever-undetected) infinite loop.

The later trait-wiring and key-generation commits (`f1a87e2`..`1efdd64`) have their own scoped mutants run in progress, covering the 14 files those commits touch; that result will follow as a separate comment once complete.

dghgit:
## cargo mutants -- scoped run over the trait-wiring + keygen commits (`f1a87e2`..`1efdd64`)

Scoped to the 14 files those commits touch (`emsa_pkcs1_v1_5.rs`, `emsa_pss.rs`, `emsa_pss_shake.rs`, `keygen.rs`, `keys.rs`, `rsa_1024/1536/2048/3072/4096/8192.rs`, `rsassa_pkcs1_v1_5.rs`, `rsassa_pss.rs`, `rsassa_pss_shake.rs`), `--jobs 3 --timeout 3000`, single-threaded tests.

**Result: 751/751 mutants -- 476 caught, 0 missed, 180 unviable, 95 timeout.** Every timeout was individually investigated (not just left as "presumably fine"); none is a silent gap. They break down as:
- **2 build timeouts** -- cold scratch-copy compile time exceeding the auto-set cap, unrelated to the mutation.
- **~14 genuine infinite loops** -- real hangs, i.e. legitimate kills, just slow ones. All in `keygen.rs`'s retry/gate logic (see below) or `has_small_factor`'s prime-sieve loop.
- **~79 equivalent-at-cap** -- the mutated code runs to completion correctly (confirmed by watching many full test binaries pass in the log) but the single-threaded suite's wall-clock is close enough to the 3000s cap that the run gets cut off deep in the 8192-bit tests before finishing everything. Not equivalents in the strict sense for all of these -- see the structural notes below -- but none change the outcome the tests check.

### Why so many timeouts turn out harmless
- **`gcd`/`div`/`shl`/`shr`/`trailing_zeros`/`mask_to_bits`/`inverse_of_e_mod` in `keygen.rs` feed *only* the FIPS 186-5 criterion-3 "`d > 2^(nlen/2)`" gate check**, never the actual signing exponents (`dP`/`dQ`/`qInv`, computed independently via Fermat's theorem). Breaking these helpers can make the gate wrongly retry (a real infinite loop, ~14 of the above) or wrongly skip a redundant margin check (harmless, since the key is still correct) -- but it can never produce a broken key.
- **`has_small_factor`'s prime-batching loop** has three distinct ways its index bookkeeping can be broken into a genuine, deterministic infinite loop (confirmed by tracing the fixed batching structure by hand, not just observing the hang).
- **`is_probable_prime`'s parity and small-value fast paths** are redundant shortcuts whose fallback (full trial division / Miller-Rabin) independently reaches the same correct answer, so breaking the shortcut is invisible.
- **`candidate()`'s odd-forcing bit** (`out[0] |= 1` mutated to `&=`) degrades the low-order entropy of generated primes without ever producing an invalid one -- rejected-if-even candidates are simply retried. Functionally harmless, but a real (if narrow) reduction in randomness quality that no functional test could ever catch.

### Four things worth your attention (none are silent mutants -- all found by manually reasoning through the timeout set, not something cargo-mutants itself flagged as missed)

1. **`primes_too_close`'s rejection mechanism is never tested against a deliberately close `p`/`q` pair.** `keygen_tests.rs` independently re-derives `|p-q| > 2^(nlen/2-100)` on whatever real key came out, rather than checking the rejection fired -- and two random ~1024-bit primes landing within that (astronomically small) distance essentially never happens by chance, so the check's disablement is invisible either way.
2. **`keygen_from_rng`'s RNG-security-strength gate is only ever tested strictly above the required strength, never exactly at it.** The suite's two RNG scenarios are `None` (must fail) and `_256bit` (must succeed) -- nothing tests an RNG at exactly the strength a given modulus size requires, which is the *ordinary* real-world case, not an edge case.
3. **Highest-leverage one: `core-test-framework`'s `test_boundary_conditions` (used by every RSA size's `test_keys` call) never asserts a real `PK::from_bytes(&pk.encode()) == pk` round trip** -- only that the encoded array has the declared (compile-time-fixed) length and that malformed-length buffers get rejected. Confirmed exhaustively: `encode`/`encode_out` are untested for every RSA size and both key-type flavors except RSA-2048 (which alone has an extra, independent round-trip test in `keys_tests.rs`). A single added assertion in the shared helper would close this for every algorithm using it, not just RSA.
4. **`rsassa_pss`'s and `rsassa_pss_shake`'s public `sign_from_hash` functions appear to be dead code.** The actual `Signer::sign_final` path calls `sign_from_hash_with_salt` directly; a crate-wide grep turns up no call site for the plain `sign_from_hash` anywhere but its own definition and doc comments. (PKCS#1-v1.5's equivalent `sign_from_hash` *is* called from its own `sign_final` -- this is PSS/PSS-SHAKE specific.) Flagging as an open question rather than a bug: could be intentional public API for one-shot, non-streaming callers, or a refactor leftover that should be wired in or removed.

A `--iterate --timeout 7200 --build-timeout 900` re-run over just the 95 timeout mutants is in progress to positively confirm the genuine-infinite-loop ones don't terminate even given much more time; a follow-up comment will report that once it completes.

## AI Usage Statement
- [x] Yes, this analysis was performed by AI

Assisted-by: Claude:claude-sonnet-5

🤖 Generated with [Claude Code](https://claude.com/claude-code)

dghgit:
## Follow-up: `--iterate --timeout 7200` re-confirmation of the 95 timeout mutants

Re-ran just the 95 timeout mutants from the scoped trait-wiring/keygen run above with a much longer cap (7200s test / 900s build) to positively confirm the genuine-infinite-loop classifications rather than relying on "didn't finish in 3000s."

**Final result: 95/95 -- 76 missed, 5 caught, 14 timeout.** All 14 remaining timeouts are the genuine infinite loops already itemized above, and every one of them ran the full 7200s (or hit the 900s build-timeout, for the one const-eval compile-time hang) without resolving -- confirming non-termination rather than a slow pass.

One worth flagging precisely because it's the interesting case: `keygen.rs:162:5` (`lt` stubbed to always return `true`) was originally observed hanging under the 3000s cap and was provisionally grouped with the other infinite loops, but flagged at the time as behaving differently from truly *deterministic* hangs like `is_zero`'s -- its corruption causes a wrapping-underflow subtraction that's genuinely input-dependent, since `gcd()` is fed a fresh random `p`/`q` on every run. This time it resolved to **caught** instead: a fast, clean panic (`mod_pow needs base < modulus`) within a second, on different random primes than the earlier run drew. That's not a contradiction -- it's the input-dependent behavior playing out the other way this time, and either outcome (hang or fast panic) is a legitimate kill, never a silent pass. So the final count is 14 confirmed genuine infinite loops rather than 15, with `162:5` moving to the caught column.

This closes out the mutation-testing investigation for `f1a87e2..1efdd64`.

